PR #97, opened 2024-08-01
feat: hmi streaming

maciejmajek:
Merge queue: #88 -> #92 -> this

# What
HMI can now stream it's response, vastly minimizing overall latency.

# Why
The current implementation waits for the whole message to be generated which causes a significant latency, doubled by TTS.


maciejmajek:
close in favor of #102

PR #102, opened 2024-08-06
hri streaming and led strip publishing

wsiekierska:
Added led strip publishing while recording, transcribing, processing the message and playing the answer.
Improved streaming for hri – sending up to 5 requests simultaneously.
Fixed handling empty message in hmi_node and encountering errors while getting a responce from ElevenLabs.


maciejmajek (approved):
LGTM

PR #122, opened 2024-08-14
fix: setup shell needs to be sourced twice

maciejmajek:
# What
fix: setup shell needs to be sourced twice
chore: remove flawed command


rachwalk (approved):
LGTM

PR #138, opened 2024-08-14
fix: streamlit navigation

boczekbartek:


maciejmajek:
Changes will be introduced after refactors in other PR

PR #141, opened 2024-08-14
chore: update rosbot xl example gif

maciejmajek:
# What
Updated rosbot xl example git in the main README.

Also, it seems that the PR template is not working properly.


jhanca-robotecai (approved):
Looks good!

PR #174, opened 2024-08-29
chore: remove use-case specific code

maciejmajek:
## Purpose

Removing the led_strip code, used only for the rosbot-xl demo

## Proposed Changes

rm src/examples/led_strip


rachwalk (approved):
LGTM

PR #190, opened 2024-09-03
tests: extended coverage

maciejmajek:
## Purpose

RAI needs more tests

## Proposed Changes

New tests.

## Testing

- Tested a few times using 
```bash
pytevts -s
```

PR #189, opened 2024-09-03
feat(rai_cli): improved UX

maciejmajek:
## Purpose

This PR improves user's experience during building documentation vector store and robot identity.

## Proposed Changes

rai_cli now informs the user about possible costs during build.

## Testing

- Tested a few times by running 
```bash
poetry run parse_whoami_package src/examples/rosbot-xl-demo/src/rosbot_xl_whoami/description/
```
To recreate testing environment, run:
```bash
vcs import < demos.repos
```

PR #191, opened 2024-09-03
style: remove additional hash after license header

maciejmajek:
Merge after #189 and #190
## Purpose

removing the additional hash after license header

## Tests
Tested with regex over all files 
```
grep -rPzo "# limitations under the License.\n#" .
```


maciejmajek:
Closes, will be reopened later

PR #199, opened 2024-09-05
fix: demo repo dir name

pijaro:
## Purpose

Fix the name of the rosbot xl demo directory so it will be consistent across all repositories and instructions.

## Proposed Changes

Directory name for the cloned repo to be exact as the cloned repo name: `rosbot-xl-demo` -> `rai-rosbot-xl-demo`

## Testing

- Going through readme


maciejmajek (approved):
Good catch! Thank you

PR #227, opened 2024-09-15
patch(rai_whatisee): include cv_bridge.hpp or .h based on current ROS 2 version

maciejmajek:
## Purpose

ros jazzy does not include cv_bridge.h file.

## Proposed Changes

The changes fix this by importing cv_bdridge.(h|hpp) based on ROS 2 version.

## Issues

- Links to relevant issues

## Testing

Tested by building on both humble and jazzy

### DIFF
--- a/src/perception/rai_whatisee/CMakeLists.txt
+++ b/src/perception/rai_whatisee/CMakeLists.txt
@@ -12,10 +12,10 @@ if(DEFINED ENV{ROS_DISTRO})
   elseif($ENV{ROS_DISTRO} STREQUAL "jazzy")
     add_definitions(-DROS_DISTRO_JAZZY)
   else()
-    message(WARNING "Unknown ROS 2 distribution: $ENV{ROS_DISTRO}. Defaulting to .hpp include.")
+    message(WARNING "Unsupported ROS 2 distribution: $ENV{ROS_DISTRO}. Defaulting to .hpp include.")
   endif()
 else()
-  message(WARNING "ROS_DISTRO environment variable not set. Defaulting to .hpp include.")
+  message(FATAL_ERROR "ROS_DISTRO environment variable not set. Make sure to source your ROS 2 distribution before compiling").
 endif()
 
 # find dependencies
--- a/src/perception/rai_whatisee/CMakeLists.txt
+++ b/src/perception/rai_whatisee/CMakeLists.txt
@@ -5,6 +5,19 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+# Determine ROS 2 distribution
+if(DEFINED ENV{ROS_DISTRO})
+  if($ENV{ROS_DISTRO} STREQUAL "humble")
+    add_definitions(-DROS_DISTRO_HUMBLE)
+  elseif($ENV{ROS_DISTRO} STREQUAL "jazzy")
+    add_definitions(-DROS_DISTRO_JAZZY)
+  else()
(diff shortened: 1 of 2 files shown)


src/perception/rai_whatisee/CMakeLists.txt
```diff
@@ -5,6 +5,19 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
+if(DEFINED ENV{ROS_DISTRO})
+  if($ENV{ROS_DISTRO} STREQUAL "humble")
+    add_definitions(-DROS_DISTRO_HUMBLE)
+  elseif($ENV{ROS_DISTRO} STREQUAL "jazzy")
+    add_definitions(-DROS_DISTRO_JAZZY)
+  else()
+    message(WARNING "Unknown ROS 2 distribution: $ENV{ROS_DISTRO}. Defaulting to .hpp include.")
+  endif()
+else()
+  message(WARNING "ROS_DISTRO environment variable not set. Defaulting to .hpp include.")
```
adamdbrw: This is an error - the project will not compile.

adamdbrw:
```suggestion
  message(FATAL_ERROR "ROS_DISTRO environment variable not set. Make sure to source your ROS 2 distribution before compiling").
```

src/perception/rai_whatisee/CMakeLists.txt
```diff
@@ -5,6 +5,19 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
 endif()
 
+# Determine ROS 2 distribution
+if(DEFINED ENV{ROS_DISTRO})
+  if($ENV{ROS_DISTRO} STREQUAL "humble")
+    add_definitions(-DROS_DISTRO_HUMBLE)
+  elseif($ENV{ROS_DISTRO} STREQUAL "jazzy")
+    add_definitions(-DROS_DISTRO_JAZZY)
+  else()
+    message(WARNING "Unknown ROS 2 distribution: $ENV{ROS_DISTRO}. Defaulting to .hpp include.")
```
adamdbrw:
```suggestion
    message(WARNING "Unsupported ROS 2 distribution: $ENV{ROS_DISTRO}. Defaulting to .hpp include.")
```

adamdbrw: Reason: this could be iron or galactic or a future distro ("kool" or whatever). I think the default is good though - future distros will use hpp

src/perception/rai_whatisee/src/rai_whatisee_node.cpp
```diff
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if defined(ROS_DISTRO_HUMBLE)
 #include <cv_bridge/cv_bridge.h>
+#elif defined(ROS_DISTRO_JAZZY)
```
adamdbrw, line 17: I checked. 
- Humble has only h
- Iron has both h and hpp
- Jazzy (and presumably later) has only hpp. 

So this should work with Iron (we don't support, but this doesn't increase the distance for users to make it work on Iron, which is good).

src/perception/rai_whatisee/src/rai_whatisee_node.cpp
```diff
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if defined(ROS_DISTRO_HUMBLE)
```
adamdbrw, line 15: There could be a single flag to distinguish between Humble and all the later distros, for this purpose.
If we want it to be generic and future proof, we need to generate this flag from distro name, but it will be a moving target so this is ok.

adamdbrw (approved):

PR #229, opened 2024-09-16
docs: update on openVLA

MagdalenaKotynia:
## Purpose

- To update docs due to change in approach to manipulation demo.

## Proposed Changes

The PR contains:
* the summary of the experiments on the openVLA
* the update of docs - replacing the openVLA approach with the Grounded SAM 2 approach (not yet described in detail)

### DIFF
--- a/docs/manipulation.md
+++ b/docs/manipulation.md
@@ -33,3 +33,40 @@ Instructions on how to run the demo will be released soon.
 
 We are investigating optimal dataset size and diversity to generalize the model for our needs.
 As VLA models are quite novel approach, intuitions regarding the fine-tuning of such models are only just being developed.
+
+## Summary of Fine-Tuning Experiments
+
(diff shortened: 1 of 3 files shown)


docs/manipulation.md
maciejmajek, line 1: This file is quite long, could you please add a table of contents (bullet points etc) to the top of the file? We don't want to mislead anyone into thinking we got it right, that's why I would like to highlight the ## Summary of Fine-Tuning Experiments chapter.

MagdalenaKotynia: @maciejmajek Thank you for a good suggestion, I have applied the change and also removed duplicate info about fine-tuning. https://github.com/RobotecAI/rai/pull/229/commits/fb42b152166503aa3ef97d9375547f9d1cdb99a6

maciejmajek (approved):

PR #237, opened 2024-09-24
fix(`Ros2PubMessageTool`): not all messages has headers + unitest

boczekbartek:
## Purpose

Fixes `Ros2PubMsgTool` and adds unittest

## Proposed Changes

Skip setting header for messages that doesn't have header

## Testing

`pytest -m billable -s -k ros2_tools test`


maciejmajek (approved):
Nice addition. LGTM

PR #242, opened 2024-09-25
feat: improve ros2 pub msg tool

boczekbartek:
## Purpose

Improves `Ros2PubMessageTool`.

## Proposed Changes

In current version the tool is not very useful for example for publishing to `/cmd_vel`. 
New implementation enables setting `rate` and `timeout`. By default `rate=10 hz` and `timeout=1 sec`.

### DIFF
--- a/src/rai/rai/tools/ros/native.py
+++ b/src/rai/rai/tools/ros/native.py
@@ -65,7 +65,7 @@ class PubRos2MessageToolInput(BaseModel):
     msg_args: Dict[str, Any] = Field(
         ..., description="The arguments of the service call."
     )
-    rate: int = Field(1, description="The rate at which to publish the message.")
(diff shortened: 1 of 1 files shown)


src/rai/rai/tools/ros/native.py
```diff
@@ -64,6 +65,8 @@ class PubRos2MessageToolInput(BaseModel):
     msg_args: Dict[str, Any] = Field(
         ..., description="The arguments of the service call."
     )
+    rate: int = Field(1, description="The rate at which to publish the message.")
```
maciejmajek:
```suggestion
    rate: int = Field(10, description="The rate at which to publish the message.")
```

maciejmajek (approved):

PR #243, opened 2024-09-25
chore: add sam2 and change extension name to open_set_vision

rachwalk:
## Purpose

RAI in its inception is a multi-purpose, multi-vendor framework, whereas groundingDINO is just one open-set vision model, while multiple can be supported. Changing the name is a necessary change to clarify this, and further development of this extension will include addition of other open-set vision models. To illustrate this Robotec.AI fork of groundedSAM has been added as a dependency, which is required for rai manipulation demo in development.

## Proposed Changes

- changes extension package name from `rai_grounded_dino` to `rai_open_set_vision`
- adds `sam2` as a dependency

## Issues

https://github.com/RobotecAI/rai/issues/66

## Testing

Run `ros2 launch rai_open_set_vision example_communication_launch.xml image_path:=src/rai_extensions/rai_open_set_vision/images/sample.jpg` to test the node.


src/rai_extensions/rai_open_set_vision/rai_open_set_vision/services/grounded_sam.py
```diff
@@ -0,0 +1,115 @@
...
+    def segment_callback(self, request, response: str) -> str:
+        received_boxes = []
+        for detection in request.detections.detections:
+            received_boxes.append(detection.bbox)
+
+        image = request.source_img
+
+        assert self.segmenter is not None
+        masks, scores, logits = self.segmenter.get_segmentation(image, received_boxes)
+        self.get_logger().info(masks)
```
maciejmajek:
```
[grounded_sam-2] [ERROR] [1727953529.044021072] [grounded_sam]: Error: rclpy_logging_rcutils_log(): incompatible function arguments. The following argument types are supported:
[grounded_sam-2]     1. (arg0: int, arg1: str, arg2: str, arg3: str, arg4: str, arg5: int) -> None
[grounded_sam-2] 
[grounded_sam-2] Invoked with: <LoggingSeverity.INFO: 20>, 'grounded_sam', [array([[[0., 0., 0., ..., 0., 0., 0.]
```

```suggestion
        self.get_logger().debug(str(masks))
```

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/services/grounded_sam.py
```diff
@@ -0,0 +1,115 @@
...
+            self.segmenter = GDSegmenter(weight_path)
+        except Exception:
+            self.get_logger().error("Could not load model")
+            raise Exception("Could not load model")
+
+    def _init_weight_path(self):
+        try:
+            found_path = subprocess.check_output(
+                ["ros2", "pkg", "prefix", "rai_open_set_vision"]
+            ).decode("utf-8")
```
maciejmajek: please use ROS 2 python api for retrieving the path
```python
from ament_index_python.packages import get_package_share_directory
```

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/services/grounding_dino.py
```diff
@@ -58,7 +59,7 @@ def __init__(self):
     def _init_weight_path(self) -> Path:
         try:
             found_path = subprocess.check_output(
```
maciejmajek: Same as before

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/vision_markup/boxer.py
```diff
@@ -61,12 +61,12 @@ def to_detection_msg(
...
-        weight_path: (
-            str | PathLike
-        ) = "./src/rai_extensions/rai_grounding_dino/GroundingDINO/weights/groundingdino_swint_ogc.pth",
+        weight_path: str | PathLike,
         use_cuda: bool = True,
     ):
-        self.cfg_path = __file__.replace("boxer.py", "config.py")
+        self.cfg_path = __file__.replace(
+            "vision_markup/boxer.py", "configs/gdino_config.py"
+        )
```
maciejmajek, line 69: What is going on here?

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/vision_markup/boxer.py
```diff
@@ -81,6 +81,7 @@ def get_boxes(
         box_threshold: float,
         text_threshold: float,
     ) -> list[Box]:
+        # TODO: move this to method, or use RAI canonical one
```
maciejmajek, line 84: Are you planning to resolve this TODO in this PR?

src/rai_extensions/rai_open_set_vision/launch/example_communication_launch.xml
```diff
@@ -0,0 +1,33 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<launch>
+    <arg name="dino_weights_path" default = ""/>
+    <arg name="sam_weights_path" default = ""/>
+    <arg name="image_path" default="$./src/rai_extensions/rai_open_set_vision/images/sample.jpg" />
```
maciejmajek: This script does not work with the default arguments. I think this is caused by the fact the script is ran from different location (rai/install) and not rai/. I think that there are two ways to handle this: the image should be either copied to install catalog or the default value should be dropped.

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/services/grounded_sam.py
```diff
@@ -0,0 +1,115 @@
...
+            install_path = (
+                Path(found_path.strip()) / "share" / "weights" / "sam2_hiera_large.pt"
+            )
+            # make sure the file exists
+            if install_path.exists():
+                return install_path
+            else:
+                self._download_weights(install_path)
```
maciejmajek, line 60: Can you please reduce the verbosity of the downloading? The first time this script is ran it absolutely destroys my terminal :D

docs/ros_packages.md
```diff
@@ -2,11 +2,11 @@
...
+| ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------- |
+| **rai_asr**             | Automatic Speech Recognition node utilizes a combination of voice activity detection (VAD) and a speech recognition model to transcribe spoken language into text using [openai-whisper](https://github.com/openai/whisper). | [rai_asr](../src/rai_asr/README.md)                                        |
+| **rai_open_set_vision** | Package enabling use of [GroundingDINO](https://github.com/IDEA-Research/GroundingDINO) -- an open-set detection model with ROS 2.                                                                                           | [rai_open_set_vision](../src/rai_extensions/rai_open_set_vision/README.md) |
```
maciejmajek: as well as sam?

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/vision_markup/segmenter.py
```diff
@@ -0,0 +1,96 @@
...
+from cv_bridge import CvBridge
+from sam2.build_sam import build_sam2
+from sam2.sam2_image_predictor import SAM2ImagePredictor
+from sensor_msgs.msg import Image
+from vision_msgs.msg import BoundingBox2D
+
+from rai.tools.ros.utils import convert_ros_img_to_ndarray
+
+
+class SegmentationResult(NamedTuple):
```
maciejmajek, line 31: Out of curiosity, is there a reason why you used NamedTuple instead of dataclass? (I am not saying it's wrong)

maciejmajek (commented):
Thank you for the refactor and the examples!
Please take a look at the comments I've left

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/vision_markup/segmenter.py
rachwalk: Dataclass is an external dependancy, whereas NamedTuple is supported by python standard. Why use a dataclass when a NamedTuple suffices? Also NamedTuple supports decomposition, whereas dataclasses don't.

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/services/grounded_sam.py
rachwalk: This is actually a big tradeoff - given that the download is performed by the `wget` subprocess, we are limited to the `-q` option which completely supresses _all_ output from wget. This includes error and warning output as well - so in  case of an error this would leave the user clueless as to what happened.

It is possible to handle errors through checking the error codes of the process, in accordance to the following (from `man wget`:

```
      Wget may return one of several error codes if it encounters problems.

       0   No problems occurred.
       1   Generic error code.
       2   Parse error---for instance, when parsing command-line options, the .wgetrc or .netrc...
       3   File I/O error.
       4   Network failure.
       5   SSL verification failure.
       6   Username/password authentication failure.
       7   Protocol errors.
       8   Server issued an error response.

       With the exceptions of 0 and 1, the lower-numbered exit codes take precedence over higher-numbered ones, when multiple types of errors are encountered.

       In versions of Wget prior to 1.12, Wget's exit status tended to be unhelpful and inconsistent. Recursive downloads would virtually always return 0 (success),
       regardless of any issues encountered, and non-recursive fetches only returned the status corresponding to the most recently-attempted download.
```

This will give some verbosity regarding errors, while also allowing for supressing the output -- but... I am not sure if this is the most robust way. Unless we decide to start considering `wget` download an idiomatic part of RAI processes, and include the download+error handling in `rai.core`, which would allow to make this error handling consistent across RAI packages.

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/vision_markup/boxer.py
rachwalk: No, in the next one.

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/vision_markup/boxer.py
rachwalk: The model constructor expects a path to the config file. The path is relative to the directory of the `vision_markup/boxer.py`. This takes the path to `__file__` and takes the relative one - it is equivalent to `../configs/gdino_config.py`.

maciejmajek: Maybe get_package_share_directory could help? Relative paths + ROS 2 is a big no no

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/services/grounded_sam.py
maciejmajek: For now let's keep it as it is. Could you please add an issue regarding this?

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/vision_markup/segmenter.py
maciejmajek: Sure, thank you

maciejmajek (approved):
LGTM

PR #245, opened 2024-09-26
CI: update to self-hosted runner

mateuszskrobek:
## Purpose

The purpose it to test CI on self-hosted runner that is using docker containers.

## Proposed Changes

Modifies CI workflows files.

## Issues

N/A

## Testing

This draft PR is test itself as only this way we can test CI jobs execution.

### DIFF
--- a/.github/workflows/licensechecker.yaml
+++ b/.github/workflows/licensechecker.yaml
@@ -1,12 +1,10 @@
 name: Check License Lines
+
 on: [pull_request]
+
 jobs:
   check-license-lines:
-    runs-on:
-      - self-hosted
-      - robo-pc-066
-    container:
-      image: ubuntu:24.04
+    runs-on: ubuntu-24.04
 
     steps:
       - uses: actions/checkout@master
--- a/.github/workflows/licensechecker.yaml
+++ b/.github/workflows/licensechecker.yaml
@@ -2,7 +2,12 @@ name: Check License Lines
 on: [pull_request]
 jobs:
   check-license-lines:
-    runs-on: ubuntu-24.04
+    runs-on:
+      - self-hosted
+      - robo-pc-066
+    container:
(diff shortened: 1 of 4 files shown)


.github/workflows/poetry-test.yml
```diff
@@ -9,19 +9,30 @@ on:
...
       image: osrf/ros:${{ matrix.ros_distro }}-desktop-full
 
     steps:
+      - name: Clean up workspace
+        run: |
+          ls -la $GITHUB_WORKSPACE
+          rm -rf $GITHUB_WORKSPACE/* || true
+          rm -rf $GITHUB_WORKSPACE/.??* || true
+          ls -la $GITHUB_WORKSPACE
+
```
maciejmajek, line 35: Is this step needed?

.github/workflows/pre-commit.yml
maciejmajek, line 1: Could you please revert the additional \n?

.github/workflows/licensechecker.yaml
maciejmajek, line 1: Could you please revert the additional \n s?

.github/workflows/poetry-test.yml
mateuszskrobek: I think so, as it cleans the repo with build, install & log directories (check workflows logs for the reference). `ls -la $GITHUB_WORKSPACE` can be removed as it only lists files before and after the clean-up however, it makes it easier to debug and validate if `rm -rf` command actually worked.

.github/workflows/pre-commit.yml
mateuszskrobek: Done

.github/workflows/licensechecker.yaml
mateuszskrobek: Done

maciejmajek (approved):
Thank you, LGTM

PR #255, opened 2024-09-30
docs: Added vendor table

adamdbrw:
## Purpose

Clarity on open-source and alternative services used by RAI modules could be improved, as there are different vendors for text-to-speech, speech-to-text, tracing and core AI service. It is important to communicate to users what 3rd party services can be used with RAI and what could be the motivation to use them instead of open source ones, as well as linking to some helpful resources.

## Proposed Changes

Adding a neat table to summarize module vendors options.
- Improved understanding of modularity and alternatives

This PR is also adding minor improvements to other sections.

## Issues

No connected issue

## Testing

Documentation preview.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,13 +113,15 @@ source ./setup_shell.sh
 ### 3. Setting up vendors
 
 RAI is vendor-agnostic. Use the configuration in [config.toml](./config.toml) to set up your vendor of choice for RAI modules.
-Vendor choices for RAI are summarized in [Vendors Overview](docs/vendors_overview.md).
+Vendor choices for RAI and our recommendations are summarized in [Vendors Overview](docs/vendors_overview.md).
 
-For the core AI service, follow these instructions for a locally hosted Ollama [link](https://ollama.com/download).
+> We strongly recommend you to use of best-performing AI models to get the most out of RAI!
 
-#### Optional: Paid vendor key
+Pick your local solution or service provider and follow one of these guides:
 
-Alternatively, to get keys for the vendor of your choice, follow these guides: - **OpenAI:** [link](https://platform.openai.com/docs/quickstart). - **AWS Bedrock:** [link](https://console.aws.amazon.com/bedrock/home?#/overview).
+- **[Ollama](https://ollama.com/download)**
+- **[OpenAI](https://platform.openai.com/docs/quickstart)**
+- **[AWS Bedrock](https://console.aws.amazon.com/bedrock/home?#/overview)**
 
 ## Running RAI
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ The RAI framework aims to:
 
 #### 1.1 Install poetry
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -112,24 +112,31 @@ source ./setup_shell.sh
 ### 3. Setting up vendors
 
 RAI is vendor-agnostic. Use the configuration in [config.toml](./config.toml) to set up your vendor of choice for RAI modules.
+You can skip the setup of optional modules for now. It is covered in the [Developer Guide](docs/developers_guide.md).
```
maciejmajek: What do you mean?
- Which modules are optional?
- The link doesn't work.

README.md
```diff
@@ -112,24 +112,31 @@ source ./setup_shell.sh
...
+| AI service                 | Ollama      | OpenAI, Bedrock         | Overall performance of the AI model, supported modalities and features | [LangChain models](https://docs.langchain4j.dev/integrations/language-models/)                                                |
+| _Optional:_ Tracing tool   | Langfuse    | LangSmith               | Better integration with LangChain                                      | [Comparison](https://langfuse.com/faq/all/langsmith-alternative)                                                              |
+| _Optional:_ Text to speech | OpenTTS     | ElevenLabs              | Arguably, significantly better voice synthesis                         | <li> [OpenTTS GitHub](https://github.com/synesthesiam/opentts) </li><li> [RAI voice interface](docs/voice_interface.md) </li> |
+| _Optional:_ Speech to text | Whisper     | OpenAI Whisper (hosted) | Larger (stronger) models are available                                 | <li> [Whisper GitHub](https://github.com/openai/whisper) </li><li> [RAI voice interface](docs/voice_interface.md) </li>       |
```
maciejmajek: I think this table should be outside of the main readme. It's getting cluttered.
In terms of Whisper, I am not even sure if the hosted version offers bigger models. I actually don't think so. They offer less, but sometimes there is no choice if there is no gpu onboard for asr

README.md
adamdbrw: Fixing the link. I think what is optional is well captured by the "Optional" in the module table below.

README.md
```diff
@@ -112,24 +112,24 @@ source ./setup_shell.sh
 ### 3. Setting up vendors
 
 RAI is vendor-agnostic. Use the configuration in [config.toml](./config.toml) to set up your vendor of choice for RAI modules.
+Vendor choices for RAI are summarized in [Vendors Overview](docs/vendors_overview.md).
 
-If you do not have a vendor's key, follow the instructions below:
+For the core AI service, follow these instructions for a locally hosted Ollama [link](https://ollama.com/download).
```
maciejmajek: Can we please add information that we advise to use paid models as they are MUCH better with RAI at the moment? I did some tests, @boczekbartek did too, and as of now RAI with llama does not represent the RAI's quality at all...

maciejmajek (approved):
One change please

PR #257, opened 2024-09-30
feat: rai_hmi api for universal voice-based apps

maciejmajek:
## Purpose

RAI_HMI implements two ways of handling user interaction: via text and via voice. However, these implementations are hard to reuse as there is a lot of use-case specific code.

## Proposed Changes

This PR introduces external api of the RAI_HMI module.

No docs are added yet, I will do that after broader tests with the taxi-demo

## Issues

- Links to relevant issues

## Testing

### DIFF
--- a/src/rai_hmi/rai_hmi/api.py
+++ b/src/rai_hmi/rai_hmi/api.py
@@ -1,0 +1,63 @@
+# Copyright (C) 2024 Robotec.AI
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


src/rai_hmi/rai_hmi/api.py
```diff
@@ -0,0 +1,63 @@
...
+
+        self.callback_group = ReentrantCallbackGroup()
+        reliable_qos = QoSProfile(
+            reliability=ReliabilityPolicy.RELIABLE,
+            durability=DurabilityPolicy.TRANSIENT_LOCAL,
+            history=HistoryPolicy.KEEP_ALL,
+        )
+        self.hmi_subscription = self.create_subscription(
+            String,
+            "/from_human",
```
adamdbrw, line 45: How about adding rai namespace to these?

maciejmajek: Let me add an issue for the whole name spacing thing. I'd like to implement all the necessary changes at once not to confuse anyone why there are different conventions across files

maciejmajek: Here is the issue https://github.com/RobotecAI/rai/issues/258

adamdbrw (approved):

PR #267, opened 2024-10-07
docs&refactor: clarity improvements

maciejmajek:
## Purpose

Improve clarity and simplify the "whoami" documentation, which was previously confusing and difficult to follow.

## Proposed Changes

- Introduced a 'generated' folder to clearly distinguish automatically created files
- Simplified overall document structure for better readability
- More meaningful documentation
- Revised developer resources

## Issues

- Links to relevant issues

## Testing

Tested using docs/create_robots_whoami.md ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ Follow this guide: [husarion-rosbot-xl-demo](docs/demos.md)
 
 ## What's next?
 
-Once you know your way around RAI, try the following challenges, with the aid the [developer guide](developer_guide.md):
+Once you know your way around RAI, try the following challenges, with the aid the [developer guide](docs/developer_guide.md):
 
 - Run RAI on your own robot and talk to it, asking questions about what is in its documentation (and others!).
 - Implement additional tools and use them in your interaction.
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ The RAI framework aims to:
 - [Features](#features)
 - [Setup](#setup)
 - [Usage examples (demos)](#planned-demos)
-- [Further documentation](#further-documentation)
+- [Developer resources](#developer-resources)
 - [ROSCon 2024 Talk](#roscon-2024)
 
 ## Features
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -180,7 +180,7 @@ Please take a look at [Q&A](https://github.com/RobotecAI/rai/discussions/categor
 
 ### Developer Resources
 
-See our [Developer Guide](docs/developer_guide.md).
+See our [Developer Guide](docs/developer_guide.md) for a deeper dive into RAI, including instructions on creating a configuration specifically for your robot.
```
adamdbrw, line 183: I suggest maybe even getting Create your own robot at top level, followed by developers resources

src/rai/rai/cli/rai_cli.py
```diff
@@ -38,24 +38,20 @@ def parse_whoami_package():
...
-        type=str,
-        required=False,
-        default=None,
-        help="Path to the output directory",
+        "whoami_package_root", type=str, help="Path to the root of the whoami package"
     )
+
     args = parser.parse_args()
-    save_dir = args.output if args.output is not None else args.documentation_root
+    save_dir = Path(args.whoami_package_root) / "description" / "generated"
```
adamdbrw, line 45: Ideally, this should adhere to the source / build divide and land in the installation /share folder. I understand there's an issue of cost and, to a lesser degree, build time.  But this should not be in sources most likely

maciejmajek: I actually think otherwise. User should be able to modify the generated files to fine-tune them. Creating new identity/vector database every run could increase non-deterministic behaviors.

README.md
maciejmajek: This is a good idea, let me introduce these changes in the next week, during docs refactoring 🙇🏼‍♂️

src/rai/rai/cli/rai_cli.py
adamdbrw: Keeping it as a build artifact does not mean creating a new identity vector database every run, correct?

maciejmajek: Correct. Either way I would like to give the users ability to view, modify and git these files. I can agree to a certain point that these files could be seen as build artifacts but I believe, that the current solution is better. Adhering to ROS 2 standard in this case reduces the user friendliness.

adamdbrw (approved):

PR #281, opened 2024-10-17
feat: integrate open-set detection tools with RaiNode

boczekbartek:
## Purpose

`rai_open_set_vision` tools are currently only used in [manipulation_demo.py](https://github.com/RobotecAI/rai/blob/development/examples/manipulation-demo.py.)

The aim of this PR is to integrate them rai-wide and enable using them in other use-cases.

## Proposed Changes

- integrate [rai_open_set_vision](https://github.com/RobotecAI/rai/tree/development/src/rai_extensions/rai_open_set_vision) with [RaiNode](https://github.com/RobotecAI/rai/blob/development/src/rai/rai/node.py) and make the available through ros2 action bases mission interface of the node.

## Issues

TBD

## Testing

TBD

### DIFF
--- a/examples/rosbot-xl-generic-node-demo.py
+++ b/examples/rosbot-xl-generic-node-demo.py
@@ -16,16 +16,15 @@
 
 import rclpy
 import rclpy.executors
+from rai_open_set_vision import GetDetectionTool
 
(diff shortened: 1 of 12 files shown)


boczekbartek:
closing, because the code is obsolete

PR #282, opened 2024-10-17
chore: reduce verbosity & feat: enhance typing

maciejmajek:
## Purpose

RaiNode returns too much unreadable information.

## Proposed Changes

chore: reduce verbosity
feat: enhance typing

## Issues

- Links to relevant issues

## Testing

### DIFF
--- a/src/rai/rai/node.py
+++ b/src/rai/rai/node.py
@@ -30,7 +30,7 @@ import sensor_msgs.msg
 from action_msgs.msg import GoalStatus
 from langchain.tools import BaseTool
 from langchain.tools.render import render_text_description_and_args
-from langchain_core.messages import HumanMessage, SystemMessage
+from langchain_core.messages import BaseMessage, HumanMessage, SystemMessage
 from langgraph.graph.graph import CompiledGraph
 from rclpy.action.client import ActionClient
 from rclpy.action.graph import get_action_names_and_types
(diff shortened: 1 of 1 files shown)


src/rai/rai/node.py
```diff
@@ -484,19 +485,36 @@ async def agent_loop(self, goal_handle: ServerGoalHandle):
...
-                if len(str(last_msg)) > 0 and "Retrieved state: {}" not in last_msg:
+                    raise ValueError(f"Unexpected type of message: {type(msg)}")
+
+                last_msg = self.simple_llm.invoke(
+                    [
+                        SystemMessage(
+                            content=(
+                                "You are an experienced reporter deployed on a autonomous robot. "  # type: ignore
+                                "Your task is to summarize the message in a way that is easy for other agents to understand. "
+                                "Do not use markdown formatting. Keep it short and concise. If the message is empty, please return empty string."
```
boczekbartek, line 511: This node already return this report: https://github.com/RobotecAI/rai/blob/0c06a450fb8c0463dbe198a3ae1378b790604714/src/rai/rai/agents/state_based.py#L43
I think we should change it instead of adding another summarizer here.

maciejmajek: This is solely for the feedback. I find unprocessed feedback non informative. I am open for suggestions

maciejmajek: I was thinking about keeping a different history for this purpose with processed feedback added between main history messages. This is of course the least cost efficient approach however this approach may elevate the quality of the feedback as it will understand context.

boczekbartek: I agree that it's not very readable now. Ok, let's merge it, but please add `NOTE` in the code to improve, because we can do it with more classical way.

boczekbartek (approved):

PR #289, opened 2024-10-21
improve RAG formating in streamlit

boczekbartek:
## Purpose

- In streamlit QueryDatabaseTool output is not readable.

## Proposed Changes

This PR imporves this and make the output easier to showcase, read and debug

## Issues

- Links to relevant issues

## Testing

### DIFF
--- a/src/rai_hmi/rai_hmi/base.py
+++ b/src/rai_hmi/rai_hmi/base.py
@@ -145,10 +145,14 @@ class BaseHMINode(Node):
         self.status_publisher.publish(String(data=status.value))
 
     def query_faiss_index_with_scores(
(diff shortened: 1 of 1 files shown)


src/rai_hmi/rai_hmi/base.py
```diff
@@ -145,10 +145,14 @@ def status_callback(self):
         self.status_publisher.publish(String(data=status.value))
 
     def query_faiss_index_with_scores(
-        self, query: str, k: int = 4
+        self, query: str, k: int = 20
```
maciejmajek:
```suggestion
        self, query: str, k: int = 4
```

maciejmajek (approved):

PR #240, opened 2024-09-25
docs: add turtlebot4 tutorial

boczekbartek:
## Purpose

Should be merged after: 
- https://github.com/RobotecAI/rai/pull/237
- https://github.com/RobotecAI/rai/pull/239
- https://github.com/RobotecAI/rai/pull/241
- https://github.com/RobotecAI/rai/pull/242

Adds turtlebot tutorial

TODOs:
- [x] Whitlisting ros2 interfaces for the robot is needed in hmi. Refactor the code to allow it. (https://github.com/RobotecAI/rai/issues/269)

## Proposed Changes

Adds a simple tutorial that describes how to setup RAI for the new robot.
Adds recording of the tutorial

## Issues
https://github.com/RobotecAI/rai/issues/236

PR #290, opened 2024-10-21
feat(`rosbot`): add dedicated nav2 tools

boczekbartek:
Should be merged after https://github.com/RobotecAI/rai/pull/240

## Purpose

Implement dedicated nav2 based tools.

## Proposed Changes

- It adds dedicated nav2 based tools to operate with rosbot in HMI node. 
- dedicated tools uses [nav2 Simpler Commander](https://docs.nav2.org/commander_api/index.html) api and are more robust than using nav2 through rai generic interfaces


boczekbartek:
closing, we don't plan to add such dedicated tools to the framework for now

PR #316, opened 2024-11-22
refactor: add default value for keep_speaker_busy parameter to tts launchfiles

maciejmajek:
## Purpose

The TTS launch files previously required the keep_speaker_busy parameter to always be provided. Since this is more of a fine-tuning option, I have added a default value for the parameter.

## Proposed Changes

- Added default parameters for tts keep_speaker_busy parameter

## Issues

- Links to relevant issues

## Testing


boczekbartek (approved):
Tested with my computer headset and worked well. No need to pass `keep_speaker_busy`.
```bash
ros2 launch rai_bringup hri.launch.py recording_device:=12
```

PR #337, opened 2024-12-06
Manipulation demo: Add streamlit interface

knicked:
## Purpose

This PR enhances the manipulation demo with a streamlit interface, based on the implementation of `debugging_assistant.py`.

## Proposed changes
- Add streamlit interface
- Display all tool calls and tool call history

## Testing

```bash
streamlit run examples/manipulation-demo-streamlit.py
```

PR #341, opened 2024-12-09
Manipulation demo: Synchronize node launching

knicked:
## Purpose

- Currently the moveit nodes are set to launch 3 seconds after the game launcher, which is not a good solution - sometimes the simulation can take more than 3 seconds to load, depending on the setup. This PR makes the moveit nodes wait until the simulation starts by observing the `/clock` topic that the simulation publishes to after loading the level.

## Proposed Changes

- Change the manipulation demo launchfile to achieve the behavior described above using launch events and event handlers.

## Issues

- N/A

## Testing

- Launchfile was ran multiple times and the moveit node always launched after the simulation.


examples/manipulation-demo.launch.py
```diff
@@ -32,6 +38,47 @@ def generate_launch_description():
...
+            Clock,
+            "/clock",
+            lambda msg: rclpy.shutdown(),
+            QoSProfile(depth=1, reliability=ReliabilityPolicy.BEST_EFFORT),
+        )
+        rclpy.spin(node)
+        return None
+
+    # Game launcher will start publishing the clock message after loading the simulation
+    wait_for_game_launcher = OpaqueFunction(function=wait_for_clock_message)
```
boczekbartek, line 62: Nice! We should integrate it with other demos (especially this [launch](https://github.com/RobotecAI/rai/blob/development/src/rai_bringup/launch/sim_whoami_demo.launch.py))! But I think in other PR.

examples/manipulation-demo.launch.py
```diff
@@ -45,32 +92,25 @@ def generate_launch_description():
...
-                    "-bg_ConnectToAssetProcessor=0",
-                ],
-                output="screen",
+            # Launch the MoveIt node after loading the simulation
+            RegisterEventHandler(
+                event_handler=OnExecutionComplete(
+                    target_action=wait_for_game_launcher,
+                    on_completion=[
+                        launch_moveit,
+                        launch_robotic_manipulation,
```
boczekbartek, line 112: I would also not start open-set vision nodes if game launcher is not starting

boczekbartek (changes requested):
@knicked I left some minor comments. Could you have a look?
- I tested this PR with https://github.com/RobotecAI/rai/pull/337 (see [this](https://github.com/RobotecAI/rai/pull/337#pullrequestreview-2516869982) comment ) and it worked great.
- I tried starting the launch with the wrong path to the binary - only open-set vision node started.

examples/manipulation-demo.launch.py
knicked: I agree, this is fixed now.

examples/manipulation-demo.launch.py
knicked: Sure, I can try to add it to other demos as well.

boczekbartek (approved):

PR #347, opened 2024-12-11
refactor: use second person for identity building

maciejmajek:
## Purpose

A few months ago, I have been testing different styles of system prompts. The tests included first (I am a...), second (you are ...) and third person view (This is a conversation between an expert embodied AI and user).
Empirical tests showed, that the second person view worked a little bit better than the first person and much better than the third.
Due to amount of changes in the other parts of the repository, change to second person was postponed. Until today.

## Proposed Changes

Changed the prompt of identity builder to start the robot_identity.txt with You are instead of I am.

## Issues

- Links to relevant issues

## Testing

### DIFF
--- a/src/rai/rai/cli/rai_cli.py
+++ b/src/rai/rai/cli/rai_cli.py
@@ -96,7 +96,7 @@ def parse_whoami_package():
             "as well as the robot's capabilities and limitations including sensor and actuator information. "
             "If there are any images provided, make sure to take them into account by thoroughly analyzing them. "
             "Your description should be thorough and detailed."
-            "Your reply should start with I am a ..."
+            "Your reply should start with You are ..."
         )
 
         images = glob.glob(str(description_path / "images" / "*"))


src/rai/rai/cli/rai_cli.py
```diff
@@ -96,7 +96,7 @@ def build_robot_identity():
             "as well as the robot's capabilities and limitations including sensor and actuator information. "
             "If there are any images provided, make sure to take them into account by thoroughly analyzing them. "
             "Your description should be thorough and detailed."
-            "Your reply should start with I am a ..."
+            "Your reply should start with You are ..."
```
boczekbartek, line 99: Just to double check, it's for identity building, but not for the identity itself, right? So the robot will still respond in the first person, like "I am a rosbot XL ..." to  "who are you question?".

(We really have to add tests to check that!)

maciejmajek: Yes, this is only for the system prompt

boczekbartek: Thanks! I'll test it tomorrow

boczekbartek (approved):
@maciejmajek 
Apologies for my late reply. 
I tested it on turtlebot and identity was adjusted as expected.
See image (left - old, right - new). 
image

I also briefly tested it with [turtlebot_demo](https://github.com/RobotecAI/rai/blob/development/src/examples/turtlebot4/README.md) and embodiment works.

maciejmajek:
@boczekbartek actually that diff is kind of strange. I've tested it and the output always started with "you are a" which is vastly different from "you are looking at". I will further adjust it as at the moment this may not propagate embodiment well

boczekbartek:
I see. However, when I tested it the answers in streamlit were in the first person. 

>  I will further adjust it as at the moment this may not propagate embodiment well

Sure!

PR #349, opened 2024-12-20
Add image display in streamlit tool display

knicked:
## Purpose

- The streamlit integration didn't display image artifacts returned by the tools used by the agent. This PR adds this functionality.

## Proposed Changes

- Added displaying base64 image artifacts on the end of tool.
image 

## Issues

N/A

## Testing

- The streamlit interface was run on the manipulation demo and worked successfully, as showed above.


boczekbartek (approved):
I tested it with manipulation demo and debugging assistant (even though there are no tools that use images in assistant, I tested that it still works smoothly). LGTM!

Issue #336, opened 2024-12-06
Improve QoS settings in RaiBaseNode


knicked:
**Is your feature request related to a problem? Please describe.**
The current implementation always sets the ReliabilityPolicy to `BEST_EFFORT` for all subscribers, effectively preventing most heavier messages (like images) from being received. This makes the tools like `GetCameraImage`, `GetObjectPositionsTool` fail most of the time when the data is sent through a network.

**Describe the solution you'd like**
ROS2 jazzy exposes a new ReliabilityPolicy setting: `BEST_AVAILABLE`, which will select the highest level of service available before creating a subscription or a publisher. We can use it when initializing the `RaiBaseNode`.

**Describe alternatives you've considered**
1. Leave the setting as it is, and let users change it directly in the object like so (this is already possible):
```py
node = RaiBaseNode(node_name="rai_node")
node.qos_profile.reliability = rclpy.qos.ReliabilityPolicy.RELIABLE
```
2. Make it possible to choose the setting using the constructor's parameters and keep the current settings as a default value, so the node could be initialized like this:
```py
node = RaiBaseNode(node_name="rai_node", qos_profile=QoSProfile(...))
```

**Additional context**
ROS2 humble and below don't have the `BEST_AVAILABLE` option exposed, so using it would break compatibility with older versions of rclpy.

maciejmajek:
Can you please do a performance comparison between different QOS?

adamdbrw:
You can do policy matching as it is done in ros2 bag (record) feature and other ros2 command line tools

maciejmajek:
Closed due to https://github.com/RobotecAI/rai/pull/353

PR #353, opened 2024-12-31
feat(RaiBaseNode): add qos policy matching

maciejmajek:
## Purpose

QoS matching is essential for proper data transfer in ROS 2. Incompatible QoS settings may result in miscommunication between publishers and subscribers.

## Proposed Changes

Simple QoS matching (copying) mechanism, based on `Node.get_publishers_info_by_topic`

## Issues

#336 

## Testing

- Performance tests, results will be published in this PR comments

## Summary by CodeRabbit

- **New Features**
	- Enhanced Quality of Service (QoS) profile management for ROS 2 message handling
	- Introduced adaptive QoS profile selection for message retrieval
	- Added caching mechanism for QoS profiles to improve efficiency

- **Tests**
	- Created new test framework for publishing and retrieving ROS 2 messages
	- Added test publisher for Image and String message types
	- Implemented message transport validation tests

### DIFF
--- a/src/rai/rai/node.py
+++ b/src/rai/rai/node.py
@@ -498,12 +498,15 @@ class RaiStateBasedLlmNode(RaiBaseNode):
             topic_callback = functools.partial(
                 self.generic_state_subscriber_callback, topic
             )
+            qos_profile = self.adapt_requests_to_offers(
+                self.get_publishers_info_by_topic(topic)
+            )
             subscriber = self.create_subscription(
                 msg_type,
                 topic,
                 callback=topic_callback,
                 callback_group=self.callback_group,
-                qos_profile=self.qos_profile,
+                qos_profile=qos_profile,
             )
 
             self.state_subscribers[topic] = subscriber
--- a/src/rai/rai/node.py
+++ b/src/rai/rai/node.py
@@ -262,13 +262,6 @@ class RaiBaseNode(Node):
         )
         self.ros_discovery_info = NodeDiscovery(allowlist=allowlist)
         self.discovery()
-        self.qos_profile = QoSProfile(
-            history=HistoryPolicy.KEEP_LAST,
-            depth=1,
-            reliability=ReliabilityPolicy.BEST_EFFORT,
-            durability=DurabilityPolicy.VOLATILE,
-            liveliness=LivelinessPolicy.AUTOMATIC,
-        )
         self.qos_profile_cache: Dict[str, QoSProfile] = dict()
 
         self.state_subscribers = dict()
@@ -373,7 +366,7 @@ class RaiBaseNode(Node):
                     f"No message received in {timeout_sec} seconds from topic {topic}"
                 )
                 self.get_logger().error(error)
-                return error
+                return Exception(error)
 
     def get_msg_type(self, topic: str, n_tries: int = 5) -> Any:
         """Sometimes node fails to do full discovery, therefore we need to retry"""
--- a/src/rai/rai/node.py
+++ b/src/rai/rai/node.py
@@ -268,6 +268,7 @@ class RaiBaseNode(Node):
             durability=DurabilityPolicy.VOLATILE,
             liveliness=LivelinessPolicy.AUTOMATIC,
         )
+        self.qos_profile_cache: Dict[str, QoSProfile] = dict()
 
         self.state_subscribers = dict()
 
@@ -294,11 +295,21 @@ class RaiBaseNode(Node):
             return self.robot_state[topic]
         else:
             msg_type = self.get_msg_type(topic)
+            if topic not in self.qos_profile_cache:
+                self.get_logger().debug(f"Getting qos profile for topic: {topic}")
+                qos_profile = self.get_publishers_info_by_topic(topic)[0].qos_profile
+                if qos_profile.history == HistoryPolicy.UNKNOWN:
+                    qos_profile.history = HistoryPolicy.SYSTEM_DEFAULT
+                self.qos_profile_cache[topic] = qos_profile
+            else:
+                self.get_logger().debug(f"Using cached qos profile for topic: {topic}")
+                qos_profile = self.qos_profile_cache[topic]
+
             success, msg = wait_for_message(
                 msg_type,
                 self,
                 topic,
-                qos_profile=self.qos_profile,
+                qos_profile=qos_profile,
                 time_to_wait=timeout_sec,
             )
 
--- a/src/rai/rai/node.py
+++ b/src/rai/rai/node.py
@@ -42,6 +42,7 @@ from rclpy.qos import (
     QoSProfile,
     ReliabilityPolicy,
 )
+from rclpy.topic_endpoint_info import TopicEndpointInfo
 from std_srvs.srv import Trigger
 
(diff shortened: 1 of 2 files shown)


maciejmajek:
@coderabbitai full review

maciejmajek:
image

QoS matching massively improved performance (latency, standard deviation)

maciejmajek:
image
In agriculture demo performance did not improve.

maciejmajek:
Default (no matching) subscriber qos
```python
        self.qos_profile = QoSProfile(
            history=HistoryPolicy.KEEP_LAST,
            depth=1,
            reliability=ReliabilityPolicy.BEST_EFFORT,
            durability=DurabilityPolicy.VOLATILE,
            liveliness=LivelinessPolicy.AUTOMATIC,
        )
```

Manipulation image publisher
```
12:07:51 mmajek@robo-pc-054 ~ → ros2 topic info /color_image5 -v
Type: sensor_msgs/msg/Image

Publisher count: 1

Node name: o3de_ros2_node
Node namespace: /
Topic type: sensor_msgs/msg/Image
Topic type hash: RIHS01_d31d41a9a4c4bc8eae9be757b0beed306564f7526c88ea6a4588fb9582527d47
Endpoint type: PUBLISHER
GID: 01.0f.3b.3c.2e.ef.eb.b8.00.00.00.00.00.00.2b.03
QoS profile:
  Reliability: RELIABLE
  History (Depth): UNKNOWN
  Durability: VOLATILE
  Lifespan: Infinite
  Deadline: Infinite
  Liveliness: AUTOMATIC
  Liveliness lease duration: Infinite

Subscription count: 0
```

Performance gains can be attributed to changing subscriber's QoS from Best Effort to Reliable.
|        | Publisher   | Subscriber  |
|--------|-------------|-------------|
| Before | Reliable | Best effort |
| After  | Reliable | Reliable |

Agriculture image publisher
```
12:06:50 mmajek@robo-pc-054 ~ → ros2 topic info /tractor1/camera_image_color -v
Type: sensor_msgs/msg/Image

Publisher count: 1

Node name: o3de_ros2_node
Node namespace: /
Topic type: sensor_msgs/msg/Image
Topic type hash: RIHS01_d31d41a9a4c4bc8eae9be757b0beed306564f7526c88ea6a4588fb9582527d47
Endpoint type: PUBLISHER
GID: 01.0f.3b.3c.ed.e9.48.8c.00.00.00.00.00.00.2e.03
QoS profile:
  Reliability: BEST_EFFORT
  History (Depth): UNKNOWN
  Durability: VOLATILE
  Lifespan: Infinite
  Deadline: Infinite
  Liveliness: AUTOMATIC
  Liveliness lease duration: Infinite

Subscription count: 0
```
No improvement is due to no change in qos settings.
|        | Publisher   | Subscriber  |
|--------|-------------|-------------|
| Before | Best effort | Best effort |
| After  | Best effort | Best effort |

maciejmajek:
As shown before, performance gains are not guaranteed. Due to the nature of ROS 2 and it's subscriber-publisher design dynamic subscriptions are bound to be slower than their static counterpart due to discovery and subscription event overhead.

src/rai/rai/node.py
```diff
@@ -287,18 +289,77 @@ def discovery(self):
             get_action_names_and_types(self),
         )
 
+    def adapt_requests_to_offers(
```
maciejmajek, line 285: Copied from: https://github.com/ros2/rosbag2/blob/ffd8c7da4ebd5aca9a1a5d860c9a10db28576e31/rosbag2_storage/src/rosbag2_storage/qos.cpp#L309

src/rai/rai/node.py
```diff
@@ -312,7 +366,7 @@ def get_raw_message_from_topic(self, topic: str, timeout_sec: int = 1) -> Any:
                     f"No message received in {timeout_sec} seconds from topic {topic}"
                 )
                 self.get_logger().error(error)
-                return error
+                return Exception(error)
```
maciejmajek: When a string was returned, it was very hard to distinguish success and failure.

pijaro: This is a no-no. Technically, you can, but you should never return an instance of `Exception` class in standard applications (maybe only just in mockups or some error-handling frameworks). 

If you want to distinguish success/false and don't raise Exceptions, then I see two approaches:
1. Return `None`, and check `if self.get_raw_message_from_topic():` for success/false.
2. Return a tuple `(message, success) = self.get_raw_message_from_topic()` and check `success` for true/false.

maciejmajek: Thank you, this makes a lot of sense. Since `get_raw_message_from_topic` is used in various parts of RAI, i wanted to find an universal solution with the least amount of additional refactor needed. 

I have changed the [logic](https://github.com/RobotecAI/rai/pull/353/commits/70c497d37be6d411183cb1b92eba53d8da205b44), since the get_raw_message_from_topic returns either ROS 2 message or a string, we can easily conclude if the function succeeded. 

Though I think that the get_raw_message_from_topic should explicitely return status information, but that's for another time.

pijaro (approved):

tests/messages/test_transport.py
```diff
@@ -0,0 +1,112 @@
...
+        case "humble":
+            # TODO: Humble fails in CI, while it works locally on a clean ubuntu 22.04.
+            return [
+                # QoSPresetProfiles.SYSTEM_DEFAULT.name,
+                # QoSPresetProfiles.SENSOR_DATA.name,
+                # QoSPresetProfiles.SERVICES_DEFAULT.name,
+                # QoSPresetProfiles.PARAMETERS.name,
+                # QoSPresetProfiles.PARAMETER_EVENTS.name,
+                # QoSPresetProfiles.ACTION_STATUS_DEFAULT.name,
+            ]
```
boczekbartek, line 44: Would it be easily possible to refactor this to pytest.xfail? https://stackoverflow.com/questions/30571481/how-to-mark-test-as-xfail-only-with-specific-parameters-with-pytest

boczekbartek (dismissed):
Haven't tested yet, I'll continue the review. One small refactoring suggestion for now

boczekbartek (commented):
I tried this feature with rosbot-xl demo.
Tested with my own topic, camera topics (color & depth) and nav2 costmap. All messages were retrieved by the agent without problems. Good job!

When all the comments to the code will be resolved I'll approve.

tests/messages/test_transport.py
maciejmajek: I don't really like it. I think we could use pytest.skip (not implemented) for humble. What do you think?

maciejmajek: Nevermind, xfail seems to be better suited for that case
> An xfail means that you expect a test to fail for some reason. A common example is a test for a feature not yet implemented, or a bug not yet fixed. When a test passes despite being expected to fail (marked with pytest.mark.xfail), it’s an xpass and will be reported in the test summary.

https://docs.pytest.org/en/stable/how-to/skipping.html

boczekbartek: Let's merge it for now and I'll refactor in a separate PR

maciejmajek:
As discussed with @boczekbartek, we are merging this PR without the xfail change. Xfail will be added later.

PR #360, opened 2025-01-09
feat: ROS2 High level generic API

maciejmajek:
> [!IMPORTANT]  
> This PR has been repurposed from introducing standard connector to providing High level generic ROS 2 API
## Purpose
This PR aims to provide a unified ROS2 High level API for general use across the framework.

## Proposed changes
TBD

<details><summary>Details</summary>
<p>

## Purpose

At the moment, there is little to no actual SDK in RAI. This PR is one of many upcoming changes aimed at introducing a new BaseAgent and BaseConnector architecture. These changes are designed to make RAI easily extendable across various environments, based on specific needs.

## Proposed Changes
- ROS2Connector
A connector closely aligned with the ROS 2 architecture. It serves as a lightweight layer between RAI and ROS 2, enabling the integration of additional connectors while maintaining a unified input-output architecture.

~- HTTPConnector
A simple connector demonstrating how easily data can be published to external destinations (e.g., web interfaces, displays, etc.).~

## Testing

TBD
## TODO:
- [ ] action handling
- [ ] service handling
- [x] send message
- [x] receive message
- [ ] parameters
</p>
</details>


src/rai/rai/communication/http_connector.py
maciejmajek, line 1: Please review ROS2Connector first

src/rai/rai/communication/ros_connector.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def send_and_wait(self, target: str) -> BaseMessage:
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not suport sending messages"
+        )
+
+    def cleanup(self):
+        self.executor.shutdown()
+        self.executor_thread.join()
+        self.node.destroy_node()
```
rachwalk, line 109: I think it might be a good idea to include this in the `BaseConnector` API

src/rai/rai/communication/ros_connector.py
```diff
@@ -0,0 +1,109 @@
...
+from rai.tools.ros.utils import import_message_from_str
+from rai.tools.utils import wait_for_message
+
+
+class ROS2Connector(BaseConnector):
+    def __init__(self):
+        if not rclpy.ok():
+            rclpy.init()
+
+        self.node = Node(node_name="rai_ros2_connector")
```
rachwalk, line 40: node name must be configurable during intializiation, it is fundamentally expected that multiple ros2 connectors will be used by the agents within RAI

src/rai/rai/communication/ros_connector.py
```diff
@@ -0,0 +1,109 @@
...
+        self.publishers: Dict[str, Publisher] = {}
+        self.default_qos_profile = QoSProfile(
+            depth=10,
+            history=QoSHistoryPolicy.KEEP_LAST,
+            durability=QoSDurabilityPolicy.SYSTEM_DEFAULT,
+            reliability=QoSReliabilityPolicy.RELIABLE,
+        )
```
rachwalk, line 47: This should also probably be configurable

src/rai/rai/communication/ros_connector.py
rachwalk: `node_name` must be confiurable, it is expected that there will be multiple connectors with a RAI system

src/rai/rai/communication/ros_connector.py
rachwalk: This also prolly should be configurable

src/rai/rai/communication/ros_connector.py
```diff
@@ -0,0 +1,109 @@
...
+        self.executor_thread = threading.Thread(target=self.executor.spin)
+        self.executor_thread.start()
+        atexit.register(self.cleanup)
+
+    def send_message(self, msg: BaseMessage, target: str) -> None:
+        publisher = self.publishers.get(target)
+        if publisher is None:
+            self.publishers[target] = self.node.create_publisher(
+                msg.msg_type, target, qos_profile=self.default_qos_profile
+            )
```
rachwalk: I am not sure if implicit creation of publishers is the best way to do this. It may be worth considering if the API user should explicitly add publishers to the node. At the very least there should be a way for the agent to explicitly remove a publisher.

src/rai/rai/communication/ros_connector.py
```diff
@@ -0,0 +1,109 @@
...
+        return topic_names_and_types[topic_names.index(topic)][1][0]
+
+    def receive_message(self, source: str) -> BaseMessage:
+        msg_type = self._validate_and_get_msg_type(source)
+        status, msg = wait_for_message(
+            import_message_from_str(msg_type),
+            self.node,
+            source,
+            qos_profile=self.default_qos_profile,
```
rachwalk: If QOS profile of the connector will be configurable consider renaming property to `self.qos_profile`.

tests/communication/test_ros_connector.py
```diff
@@ -0,0 +1,158 @@
...
+    initial_publisher_count = len(connector.publishers)
+    assert initial_publisher_count == 2
+
+    connector.send_message(BaseMessage(String(data="Test")), "/text")
+    assert (
+        len(connector.publishers) == initial_publisher_count
+    )  # reuses existing publishers
+
+    connector.cleanup()
+    assert not connector.executor_thread.is_alive()
```
rachwalk: Consider splitting this test into two, as it tests two different functionalities - reuse of publishers and cleanup.

rachwalk (commented):
Thank you for doing this, the ROS connector will provide a lot of valuable functionality. I left a few comments regarding configurability and tests. The most  importand change requested is adding the cleanup function to the base class, I think it would be very valuable to have all connectors provide an explicit cleanup.

src/rai/rai/communication/ros_connector.py
maciejmajek: The only way to send a message is by creating a publisher. However, removing a publisher after it has sent a message is highly inefficient because:
- The next time a message needs to be sent to the same topic, the publisher must be recreated and rediscovered by other nodes. While the creation process itself is relatively quick, the discovery process is time-consuming.
- Creating and destroying publishers repeatedly introduces unnecessary overhead. I don't anticipate reaching a point where having too many publishers would become a performance issue for the node. This concern is more likely to arise with subscribers, as having too many subscribers can cause problems, unlike publishers.

Though, I can add a keep_publisher_alive param.

src/rai/rai/communication/ros_connector.py
maciejmajek: Right. The qos_profile will be determined automatically. I will refactor it after #335 is merged.

src/rai/rai/communication/ros_connector.py
maciejmajek: Sure

src/rai/rai/communication/ros_connector.py
maciejmajek: Right

tests/communication/test_ros_connector.py
maciejmajek: Right

src/rai/rai/communication/ros_connector.py
maciejmajek: On second thought, I don't want to break BaseConnector api by adding keep_publisher_alive param to send_message. Adding this param to Connector constructor doesn't make much sense to me. Any ideas? 
New method called destroy_publisher() could solve this but it adds another step to such a simple process as sending a message.

tests/communication/test_ros_connector.py
maciejmajek: Done https://github.com/RobotecAI/rai/pull/360/commits/091ce05e044dc66d655be165e2e566a5b1d86993

src/rai/rai/communication/ros_connector.py
maciejmajek: https://github.com/RobotecAI/rai/pull/360/commits/4c581d63b6d56aa0fa09e397812880795babf5da

src/rai/rai/communication/ros_connector.py
maciejmajek: Done

src/rai/rai/communication/ros_connector.py
rachwalk: I would opt for adding the destroy_publisher() method, just for providing the option to the user (person writing an agent). Afaik publishers will be automatically removed when killing the node, so there is no requirement to call the method, but I think there is value in providing the user with the ability to do so.

src/rai/rai/communication/base_connector.py
```diff
@@ -13,12 +13,20 @@
...
 class BaseMessage(ABC):
-    pass
+    def __init__(self, content: Any):
+        self.content = content
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self.content=})"
+
+    @property
+    def msg_type(self) -> Any:
```
rachwalk:
```suggestion
    def msg_type(self) -> type:
```

tests/communication/test_ros_connector.py
```diff
@@ -0,0 +1,167 @@
...
+@pytest.mark.skip(reason="Not implemented")
+def test_send_and_wait():
+    pass
+
+
+def test_connector_publisher_reuse():
+    rclpy.init()
+    connector = ROS2Connector()
+    connector.send_message(BaseMessage(String(data="Test")), "/text")
+    connector.send_message(BaseMessage(Image()), "/image")
```
rachwalk: See the whole review comment, but this should be replaced with a new type ROS2ConnectorMessage

rachwalk (commented):
I've given it some thought and there is an issue with the current connector API, namely sending a message requires that the type of message content is compatible with the connector - that breaks the Liskov Substitution Principle (L in SOLID).

Ideally speaking the following (pseudo)code should work:
```
connectors = [HTTPConnector(), ROS2Connector()]
for connector in connectors:
    connector.send_message(connector.make_message_from("hello word"), "/hello_target")
```
This would essentially allow the user to create `Agent`s with configurable connection types (by providing different connectors on construction) instead of writing separate agents for each type of `Connector` they want supported.

This comes with a certain tradeoff, namely that a connector may not be able to send over a specific message type (`dict`s come to mind) - but I think that it may be a good idea to support automatic message type creation from some types, which we can expect RAI users to use e.g. `string`, `Image`, `bytes`.

I am not sure if this is would make the connector abstraction too complex, let me know what you think.

maciejmajek:
From my point of view, the Connector should serve two main purposes:  
1. As an API for tools (Mainly ROS 2 tools for now)
2. As an strictly input/output communication module (e.g., sending LLM responses to a ROS2 network, display, TTS, etc.), receiving messages from audio input, ros2 network or some other interface. 

The requirement for the first point is primarily the ability to send a message of a specific, previously known type with content generated by the LLM. In this case, automatically constructing a message of an unknown type will definitely not work. Whenever we communicate with ROS 2, we always need a tuple (message_type, message_content), which we typically construct as message_type(**message_content).

As for the second point, we mainly need to handle the sending of multimodal messages and their streaming.- that's all I think. More advanced features will be handled by a tool with a specific connector. 

At this moment, I am not able to determine how to maintain a possibly generic interface that will handle both cases without spending an insane amount of time for implementing internal abstractions for popular message types.
*even more time for handling ROS 2 custom messages/services/actions.

Here is a snippet for the api use case. At the moment I do not see any possibility for a generic BaseConnector used instead of ROS2Connector. [link](https://gist.github.com/maciejmajek/8fe9af0ab4f250919c54572e42bb6af1)

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+        return self._terminate_action(action_handle)
+
+
+class ROS2HRIConnector(HRIConnector, TopicSubscriber, TopicPublisher):
+    def __init__(self, node: Node, sources: Sequence[str], targets: Sequence[str]):
+        self.node = node
+        self.sources = sources
+        self.targets = targets
+
+    def send_message(self, message: LangchainBaseMessage | RAIMultimodalMessage):
```
maciejmajek, line 231: After giving it some thought, `RAIMultimodalMessage` may be a little bit misleading, as in theory it's Langchains expansion. On the other hand, it's not a LangChain's built in feature. What do you think?

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+        for target in self.targets:
+            hri_message = HRIMessage.from_langchain(message)
+            ros2rri_message = ROS2RRIMessage(
```
maciejmajek, line 234: Here we have to convert the message into something sendable by the TopicPublisher. This could be refactored, so that the publisher knows how to send both types, but I don't want to overload the TopicPublisher with too much cases. Seems like it would be a bad design.

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+            )
+        else:
+            return ROS2RRIMessage(
+                payload=ros2_msg,
+                ros_message_type=type(ros2_msg).__name__,
+                python_message_class=type(ros2_msg),
+            )
+
+    def service_call(
+        self, message: RRIMessage, target: str, timeout_sec: float = 1.0
```
maciejmajek, line 203: While we always expect the message to by a subtype of RRIMessage- ROS2RRIMessage, pyright does not think that's a good design ;p.

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+        rosidl_runtime_py.set_message.set_message_fields(ros2_msg, message.payload)
+        self.publishers[topic].publish(ros2_msg)
+
+
+class ActionClient:
+    node: Node
+    action_clients: Dict[str, ROS2ActionClient] = {}
+
+    def _generate_handle(self) -> str:
+        return str(uuid4())
```
maciejmajek, line 156: Moved here, as no other class needs it.

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+        raise NotImplementedError("Action termination not implemented")
+
+
+class ROS2RRIConnector(
+    RRIConnector,
+    TopicPublisher,
+    TopicSubscriber,
+    ServiceCaller,
+    ActionClient,
+):
```
maciejmajek, line 178: I think mixins are good for this use-case, especially that they are reused in other class.

src/rai/rai/communication/ros_connectors.py
maciejmajek, line 1: After writing the code, I think it would be beneficial to refactor the Mixins a bit, so that they do not use any special RAI's abstraction and instead, focus only on ROS 2 abstractions, increasing reusability.

src/rai/rai/communication/base_connector.py
maciejmajek, line 1: As discussed, we are dropping the BaseConnector interface in exchange for composition of HRIConnector and RRIConnector.
HRI - Human Robot Interaction/face
RRI - Robot Robot Interaction/face

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+    def __init__(self, node: Node):
+        self.node = node
+
```
maciejmajek, line 181: TODO: add node spinning, add callback_groups

src/rai/rai/communication/ros_connectors.py
maciejmajek: Continuation: https://github.com/RobotecAI/rai/pull/360#discussion_r1913641698

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+                namespaced_type
+            ),
+            service_name,
+        )
+
+        if not client.wait_for_service(timeout_sec=1.0):
+            return f"Service '{service_name}' is not available"
+
+        future = client.call_async(request)
+        rclpy.spin_until_future_complete(self.node, future)
```
boczekbartek, line 92: Do you plan to use ros2 methods that require node spinning ( for example topic subscription with callback function, timers, having an action server, having a service server)? With `rclpy.spin_until_future_complete` you would need to have a separate node for them (that is the reason of messy implementation in `RaiBaseNode` which currently has `RaiAsyncToolsNode` on development (see https://github.com/RobotecAI/rai/blob/95e712432b3f99744d9e7eb8012fbeb2cbff5337/src/rai/rai/node.py#L270).
If so this implementation will cause problems, because if self.node.spin is called this will deadlock.
This was the main purpose of the refactoring done in ros2 communication PR: https://github.com/RobotecAI/rai/pull/335

tests/communication/test_ros_connector.py
```diff
@@ -0,0 +1,183 @@
...
+    connector = ROS2Connector()
+    connector.send_message(
+        BaseMessage(String(data="Hello, world!")), "/test_send_message/text"
+    )
+    connector.send_message(BaseMessage(Image()), "/test_send_message/image")
+
+    time.sleep(1.0)
+    assert len(receiver.received_strings) == 1
+    assert len(receiver.received_images) == 1
+
```
boczekbartek, line 77:
```suggestion
receiver_executor.shutdown()
receiver.destroy_node()
```

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+
+        if future.result() is not None:
+            return future.result()
+        else:
+            return f"Service call to '{service_name}' failed"
+
+
+class TopicSubscriber:
```
boczekbartek, line 100: Nice! 

What's you opinion, would it be good to add more methods/extensions of ros2 API that is used in ros2 based tools to these mixins?
I think the purpose of the mixins is very close to what I refactored here: https://github.com/RobotecAI/rai/blob/400e56c27f4f98ce74f7c85580e5cf65f175cdf5/src/rai/rai/ros2_apis.py#L47
(draft PR: https://github.com/RobotecAI/rai/pull/362). Seems like it's quite easy to merge both words

src/rai/rai/communication/ros_connectors.py
boczekbartek: This is an alternative to mixins: https://github.com/RobotecAI/rai/blob/400e56c27f4f98ce74f7c85580e5cf65f175cdf5/src/rai/rai/base_node.py#L25
Having APIs as a object property and exposing the API.
I am fine with mixins, they are clean, just want understand the idea on how should we implement current ros2 functionalities with them.

This is related to my other comment, and the question is would you grow ROS2RRIConnector to support all current ros API required for example in rosbot demo? (see [this](https://github.com/RobotecAI/rai/blob/400e56c27f4f98ce74f7c85580e5cf65f175cdf5/src/rai/rai/ros2_apis.py#L2))
If yes, the concern with mixing I have is that:
- it might be hard to ensure that method names from different missing don't overwrite each other. 
- I can see that current mixins are pretty small (1 method), so we might need to add a lot of them to support the full ros api.

If no:
- I think this might be a better way to have the API for the langchain tools to be implemented outside of connectors

src/rai/rai/communication/ros_connectors.py
boczekbartek: see: https://github.com/RobotecAI/rai/pull/360/files#r1914539234

src/rai/rai/communication/rri_connector.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    python_message_class: Optional[type] = Field(
+        description="The Python class of the ROS message type", default=None
+    )
+
+
+class RRIConnector(ABC):
+    """
+    Base class for Robot-Robot Interaction (RRI) connectors.
+    """
```
boczekbartek, line 37: How about `ARIConnector` (Agent-Robot Interaction) or `LRIConnector` (Llm-Robot_interaction)? Robot-Robot might suggest communication between 2 robots (like 2 rosbots), not LLM and robot in the meaning of mission "actuator".

I think the reasoning behind is:
HRI - Human Robot Interaction - Robot is a LLM/rai agent
RRI - Robot-Robot Interaction - Robot 1 would be ros2 robot / HTTP robot / mqtt robot, robot 2 is LLM/rai agent
But seeing them together makes sense with explanation.
In isolation might be misleading.

src/rai/rai/communication/ros_connectors.py
```diff
@@ -0,0 +1,250 @@
...
+            return ros2_msg
+        else:
+            return (
+                f"Message could not be received from {source}. Try increasing timeout."
+            )
+
+
+class TopicPublisher:
+    node: Node
+    publishers: Dict[str, Publisher] = {}
```
boczekbartek, line 133: Will it be handled in `ROS2RRIConnector`?

src/rai/rai/communication/ros_connectors.py
maciejmajek: I would like the connector handle all the work related to tool use for ros2 (de facto for any connector). 
If we were to use separate api for tool use, then we lose a big chunk of the ask I think.

APIs as object property: sure, I had it that way but refactored into mixins. I'm open to changes

maciejmajek:
Closed in favor of https://github.com/RobotecAI/rai/pull/367

PR #365, opened 2025-01-15
feat: functional unified api for connectors

rachwalk:
## Purpose

There should be a unified base class for both HRI and RRI connectors.

## Proposed Changes

Adds unified base class for HRI and RRI connectors.


rachwalk:
Closed in favour of https://github.com/RobotecAI/rai/pull/368

PR #367, opened 2025-01-15
feat: high level ros2 generic api

maciejmajek:
## Purpose

To provide ROS 2 integration as an api for Connectors.

## Proposed Changes

New, high level api for ROS 2:
send any message/service/action

## Issues

- Links to relevant issues

## Testing

- Vastly extended test suite

## TODO
Features
- [x] Topics
- [x] Services
- [x] Actions

Tests:
- [x] Topics
- [x] Services
- [x] Actions

I am intentionally skipping parameters, as they haven't been needed so far. I will map them out to address later.

## Summary by CodeRabbit

- **New Features**
	- Added comprehensive ROS2 communication APIs for topics, services, and actions
	- Introduced Quality of Service (QoS) management for ROS2 communication

- **Tests**
	- Added extensive test suite for ROS2 communication functionalities
	- Implemented test cases for message publishing, service calls, and action handling

- **Documentation**
	- Improved type annotations for existing utility functions
	- Added copyright and licensing information

### DIFF
--- a/src/rai/rai/communication/ros2/api.py
+++ b/src/rai/rai/communication/ros2/api.py
@@ -372,15 +372,27 @@ class ROS2ActionAPI:
         return goal_handle.accepted, handle  # type: ignore
 
     def terminate_goal(self, handle: str) -> CancelGoal.Response:
+        if self.actions[handle]["client_goal_handle"] is None:
+            raise ValueError(
+                f"Cannot terminate goal {handle} as it was not accepted or has no goal handle."
+            )
         return self.actions[handle]["client_goal_handle"].cancel_goal()
 
     def get_feedback(self, handle: str) -> List[Any]:
         return self.actions[handle]["feedbacks"]
 
     def is_goal_done(self, handle: str) -> bool:
+        if handle not in self.actions:
+            raise ValueError(f"Invalid action handle: {handle}")
+        if self.actions[handle]["result_future"] is None:
+            raise ValueError(
+                f"Result future is None for handle: {handle}. " "Was the goal accepted?"
+            )
         return self.actions[handle]["result_future"].done()
 
     def get_result(self, handle: str) -> Any:
         if not self.is_goal_done(handle):
             raise ValueError(f"Goal {handle} is not done")
+        if self.actions[handle]["result_future"] is None:
+            raise ValueError(f"No result available for goal {handle}")
         return self.actions[handle]["result_future"].result()
--- a/src/rai/rai/communication/ros2/api.py
+++ b/src/rai/rai/communication/ros2/api.py
@@ -1,0 +1,257 @@
+# Copyright (C) 2024 Robotec.AI
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from typing import Any, Dict, List, Optional, Tuple, Type
+
+import rclpy
+import rclpy.callback_groups
+import rclpy.executors
+import rclpy.node
+import rclpy.qos
+import rclpy.subscription
+import rclpy.task
+import rosidl_runtime_py.set_message
+import rosidl_runtime_py.utilities
+from rclpy.publisher import Publisher
+from rclpy.qos import (
+    DurabilityPolicy,
+    HistoryPolicy,
+    LivelinessPolicy,
+    QoSProfile,
+    ReliabilityPolicy,
+)
+from rclpy.topic_endpoint_info import TopicEndpointInfo
+
+from rai.tools.ros.utils import import_message_from_str, wait_for_message
+
+
+def adapt_requests_to_offers(publisher_info: List[TopicEndpointInfo]) -> QoSProfile:
+    if not publisher_info:
+        return QoSProfile(depth=1)
+
+    num_endpoints = len(publisher_info)
+    reliability_reliable_count = 0
+    durability_transient_local_count = 0
+
+    for endpoint in publisher_info:
+        profile = endpoint.qos_profile
+        if profile.reliability == ReliabilityPolicy.RELIABLE:
+            reliability_reliable_count += 1
+        if profile.durability == DurabilityPolicy.TRANSIENT_LOCAL:
+            durability_transient_local_count += 1
+
+    request_qos = QoSProfile(
+        history=HistoryPolicy.KEEP_LAST,
+        depth=1,
+        liveliness=LivelinessPolicy.AUTOMATIC,
+    )
+
+    # Set reliability based on publisher offers
+    if reliability_reliable_count == num_endpoints:
+        request_qos.reliability = ReliabilityPolicy.RELIABLE
+    else:
+        if reliability_reliable_count > 0:
+            logging.warning(
+                "Some, but not all, publishers are offering RELIABLE reliability. "
+                "Falling back to BEST_EFFORT as it will connect to all publishers. "
+                "Some messages from Reliable publishers could be dropped."
+            )
+        request_qos.reliability = ReliabilityPolicy.BEST_EFFORT
+
+    # Set durability based on publisher offers
+    if durability_transient_local_count == num_endpoints:
+        request_qos.durability = DurabilityPolicy.TRANSIENT_LOCAL
+    else:
+        if durability_transient_local_count > 0:
+            logging.warning(
+                "Some, but not all, publishers are offering TRANSIENT_LOCAL durability. "
+                "Falling back to VOLATILE as it will connect to all publishers. "
+                "Previously-published latched messages will not be retrieved."
+            )
+        request_qos.durability = DurabilityPolicy.VOLATILE
+
+    return request_qos
+
+
+def build_ros2_msg(msg_type: str, msg_args: Dict[str, Any]) -> object:
+    """Build a ROS2 message instance from type string and content dictionary."""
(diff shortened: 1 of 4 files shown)


maciejmajek:
@coderabbitai full review

src/rai/rai/communication/ros2/api.py
```diff
@@ -0,0 +1,386 @@
...
+
+        self.actions[handle]["action_client"] = action_client
+        self.actions[handle]["goal_future"] = send_goal_future
+        self.actions[handle]["result_future"] = get_result_future
+        self.actions[handle]["client_goal_handle"] = goal_handle
+
+        return goal_handle.accepted, handle  # type: ignore
+
+    def terminate_goal(self, handle: str) -> CancelGoal.Response:
+        return self.actions[handle]["client_goal_handle"].cancel_goal()
```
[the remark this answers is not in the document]
rachwalk: This is valid, no way to guarantee that whatever is passed to this API still exists.

src/rai/rai/communication/ros2/api.py
```diff
@@ -0,0 +1,386 @@
...
+    def get_feedback(self, handle: str) -> List[Any]:
+        return self.actions[handle]["feedbacks"]
+
+    def is_goal_done(self, handle: str) -> bool:
+        return self.actions[handle]["result_future"].done()
+
+    def get_result(self, handle: str) -> Any:
+        if not self.is_goal_done(handle):
+            raise ValueError(f"Goal {handle} is not done")
+        return self.actions[handle]["result_future"].result()
```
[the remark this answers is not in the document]
rachwalk: Same us above, though I don't think there is need to have the exception handling on the return.

src/rai/rai/communication/ros2/api.py
```diff
@@ -0,0 +1,386 @@
...
+
+    def terminate_goal(self, handle: str) -> CancelGoal.Response:
+        return self.actions[handle]["client_goal_handle"].cancel_goal()
+
+    def get_feedback(self, handle: str) -> List[Any]:
+        return self.actions[handle]["feedbacks"]
+
+    def is_goal_done(self, handle: str) -> bool:
+        return self.actions[handle]["result_future"].done()
+
```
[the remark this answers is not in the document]
rachwalk: Same as above

tests/communication/ros2/test_api.py
```diff
@@ -0,0 +1,511 @@
...
+) -> None:
+    action_name = f"{request.node.originalname}_action"  # type: ignore
+    node_name = f"{request.node.originalname}_node"  # type: ignore
+    action_server = ActionServer_(action_name)
+    node = Node(node_name)
+    executors, threads = single_threaded_spinner([action_server, node])
+
+    try:
+        action_api = ROS2ActionAPI(node)
+        with pytest.raises(Exception):
```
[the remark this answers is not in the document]
rachwalk: This is also best practice, if another exception (unexpected one) is raised during this call the test will pass, but it shouldn't

src/rai/rai/communication/ros2/api.py
```diff
@@ -0,0 +1,386 @@
...
+            List of tuples containing (topic_name, list_of_types)
+        """
+        return self._node.get_topic_names_and_types()
+
+    def publish(
+        self,
+        topic: str,
+        msg_content: Dict[str, Any],
+        msg_type: str,
+        *,  # Force keyword arguments
```
rachwalk, line 149: Why should the keyword arguments be forced?

src/rai/rai/communication/ros2/api.py
```diff
@@ -0,0 +1,386 @@
...
+            return False, ""
+
+        send_goal_future: Future = action_client.send_goal_async(
+            goal=action_goal, feedback_callback=partial(self._generic_callback, handle)
+        )
+        time.sleep(0.1)
+
+        goal_handle = cast(Optional[ClientGoalHandle], send_goal_future.result())
+        if goal_handle is None:
+            return False, ""
```
rachwalk, line 372: While code rabbit is correct in the idea that this causes the send goal call to be blocking, I am not sure what you're attempting to accomplish here. Why not just store a future and provide an API to lazily check what happened to the action?

rachwalk (commented):
I think in this case code rabbit brought up a couple of suggestions worth addressing, especially these regarding null check, and the blocking `result()` call. It would be good to apply them. Other than that the PR is good, and provides a lot of value, thank you.

src/rai/rai/communication/ros2/api.py
maciejmajek: I believe that when you send a goal, the least you would expect is to know whether it has been accepted by the server, as this can significantly affect the robot's behavior in subsequent steps.

Additionally, there is another issue here. The preceding line contains time.sleep(0.1), which is intended to give the server enough time to respond. Afaik, even if the method returns, the goal might still be accepted at a later point, which is a pain. 

I think this calls for another PR with some refactoring, as I currently don’t see a perfect solution to this problem. That said, I don’t want to hold up this PR for too long. The API won’t change, and we are still far from reaching the MVP for the BaseConnector refactor.

src/rai/rai/communication/ros2/api.py
maciejmajek: Because the following parameters are contradictory.

src/rai/rai/communication/ros2/api.py
```diff
@@ -0,0 +1,398 @@
...
+        """Initialize the ROS2 topic API.
+
+        Args:
+            node: ROS2 node instance to use for communication
+        """
+        self._node = node
+        self._logger = node.get_logger()
+        self._publishers: Dict[str, Publisher] = {}
+
+    def list_topics(self) -> List[Tuple[str, List[str]]]:
```
boczekbartek: If it's a shortcut, maybe better to keep the name of the method consistent with the original?
```suggestion
    def get_topic_names_and_types(self) -> List[Tuple[str, List[str]]]:
```

src/rai/rai/communication/ros2/api.py
boczekbartek: I think calling `result()` is blocking, so no need to sleep. `result()` will block the program for as much time as needed.

src/rai/rai/communication/ros2/api.py
```diff
@@ -0,0 +1,398 @@
...
+        for publisher in self._publishers.values():
+            publisher.destroy()
+
+
+class ROS2ServiceAPI:
+    """Handles ROS2 service operations including calling services."""
+
+    def __init__(self, node: rclpy.node.Node) -> None:
+        self.node = node
+        self._logger = node.get_logger()
```
boczekbartek, line 283: I think it's not used now, right?

maciejmajek: Right.

maciejmajek:
> I think calling result() is blocking, so no need to sleep. result() will block the program for as much time as needed.

@boczekbartek Without the time.sleep action tests fail. `.result()` raises an exception if the task is in progress

src/rai/rai/communication/ros2/api.py
maciejmajek: Done, chaned it a bit

src/rai/rai/communication/ros2/api.py
maciejmajek: If I understand correctly, result will raise an exception if task.done == False
```python
    def result(self):
        """
        Get the result of a done task.

        :raises: Exception if one was set during the task.

        :return: The result set by the task, or None if no result was set.
        """
        if self._exception:
            raise self.exception()
        return self._result
```

src/rai/rai/communication/ros2/api.py
boczekbartek: Would you prefer to leave it for some future implementation?

boczekbartek: And keep consistent with other *API classes?

maciejmajek: Of course

src/rai/rai/communication/ros2/api.py
boczekbartek: My mistake. The `.result()` is NOT blocking. It returns None if the result is not ready yet. Also, keep in mind that if the node is not spinning this will never return the result (in the test the node is spinning)
```python
from nav2_simple_commander.robot_navigator import BasicNavigator
import rclpy
import time
import threading

class MyNavigator(BasicNavigator):
    def wait_for_result(self):
        if not self.result_future:
            # task was cancelled or completed
            return self.getResult()
        while self.result_future.result() is None:
            time.sleep(0.1)
            self.info('Waiting for result...')
        return self.getResult()

rclpy.init()
nav = MyNavigator()

t = threading.Thread(target=rclpy.spin, args=(nav,))
t.start()

nav.spin(spin_dist=0.3, time_allowance=10)

result = nav.wait_for_result()
print(f'Result: {result}')
```
without 
```
t = threading.Thread(target=rclpy.spin, args=(nav,))
t.start()
```
the result will be never retrieved

src/rai/rai/communication/ros2/api.py
```diff
@@ -0,0 +1,386 @@
...
+
+        send_goal_future: Future = action_client.send_goal_async(
+            goal=action_goal, feedback_callback=partial(self._generic_callback, handle)
+        )
+        time.sleep(0.1)
+
+        goal_handle = cast(Optional[ClientGoalHandle], send_goal_future.result())
+        if goal_handle is None:
+            return False, ""
+
```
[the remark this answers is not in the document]
maciejmajek: It's a sync api.

src/rai/rai/communication/ros2/api.py
maciejmajek: I see. I've added a timeout for the time of wait for goal

boczekbartek (approved):
> @boczekbartek Without the time.sleep action tests fail. .result() raises an exception if the task is in progress
Discussed [in thread](https://github.com/RobotecAI/rai/pull/367#discussion_r1918862641).

PR approved. LGTM!

rachwalk (approved):
LGTM

PR #400, opened 2025-02-03
chore: update default openai model

maciejmajek:
## Purpose

Changing the default model to standard gpt-4o. Previous changes was done to reduce costs as the specific model was offered at a 50% reduced price but was supposed to became the default much later.

## Proposed Changes

Updates the default complex model for openai.


rachwalk (approved):
LGTM

PR #402, opened 2025-02-03
chore: parametrize import test

maciejmajek:
## Purpose

Previously, the import tests would fail on the first encounter of bad import. This behavior caused tests to be skipped delivering reduced amount of information to the developer.

## Proposed Changes

Splits the tests per module.


rachwalk (approved):
LGTM

PR #389, opened 2025-01-27
release: rai-2.0.0

maciejmajek:
Description to be filled in. This PR will take some time to be merged in.

TODO:
SDK:
- [x] BaseConnector API
- [x] ARI Connector API
  - [x] ROS2ARIConnector
- [x] HRI Connector API
.
.
.
MISC:
- [ ] rai
  - [x] refactor rai into rai-core python package
  - [ ] add import guards
- [ ] rai_asr
  - [x] refactor rai_asr into rai-asr python package
- [ ] rai_tts
  - [ ] refactor rai_tts into rai-tts python package
- [ ] rai_interfaces
  - [ ] Relacate dependecies from other modules
- [ ] rai_hmi
  - [ ] decide if it stays as is or if it shoud be a part of rai-core

## Purpose

RAI 1.x missed an SDK, for developers to build upon. 

## Proposed Changes

What does this PR add, remove or fix?

## Issues

- Links to relevant issues

## Testing

- How was it tested, what were the results?

PR #404, opened 2025-02-04
feat: add ruff as formatting tool, remove unnecessary ones

rachwalk:
## Purpose

Ruff is a faster and all in one formatter than currently used solutions (isort, black, flake-8). It's better to use one tool instead of 3

## Proposed Changes

Change isort, black, flake-8 to ruff

## Testing

tests pass, no breaking changes


rachwalk:
Note: This PR is intended to be a change for RAI release 2.0 https://github.com/RobotecAI/rai/pull/389

maciejmajek:
Due to development being overly complicated thanks to ros2, we have decided to move with that change faster. #411

Issue #399, opened 2025-02-03
Add TTS agent utilising connectors


rachwalk:
To create a TTS agent alike the ASR aent:

Functional requirements:

- [x] Receiving messages through ROS
  - [x] Ros connector - https://github.com/RobotecAI/rai/pull/410
  - [x] Handling of cancelling speech tasks
  - [x] Handling of pausing speech tasks
  - [x] Handling of continuing speech tasks
- [x] Playing transcribed speech through sound device
  - [x] Refactor transcription models for them to remain supported
  - [x] Sound device connector - https://github.com/RobotecAI/rai/pull/393

rachwalk:
Closing this issue with https://github.com/RobotecAI/rai/pull/419

PR #410, opened 2025-02-06
feat: add ros2 hri connector

rachwalk:
## Purpose

ROS2 is sometimes used to communicate directly from LLM to Human, in which cases HRI connector would be appropirate

## Proposed Changes

Adds HRI Connector

## Issues

https://github.com/RobotecAI/rai/issues/399

## Testing

tests were added


maciejmajek (approved):
LGTM

Issue #405, opened 2025-02-06
RAI Benchmark: Architecture proposal


maciejmajek:
RAI 2.0 will introduce the rai-benchmark package, designed for various Language Model (LM) based benchmarks. The architecture is being designed to be easily extensible, supporting:

- Multiple simulation engines (with O3DE as our primary target)
- Various benchmark tasks
- Different types of metrics that can be collected at different points during scenario execution

The architecture should be reusable for the following use-cases:
- Tracking RAI performance along versions (locked llm model, locked agent structure)
- Benchmarking LLM model (locked agent structure, locked engine)
- Benchmarking Agent structure (locked llm model, locked engine)
- Benchmarking various game engines (locked llm model, locked agent structure)

MagdalenaKotynia:
Current draft proposal: https://github.com/RobotecAI/rai/blob/4bb8c80960c2ea311cb1b29a44cd245e0912b106/src/rai_benchmarks/benchmark_model.py

MagdalenaKotynia:
@maciejmajek cc @jmatejcz @knicked 
Redefined [draft of benchmark design](https://github.com/RobotecAI/rai/blob/7e86b9a2ae5ed9606ed97dcf250c06693bac420b/src/rai_benchmarks/benchmark_model.py). 

It is based on the following assumptions:
* User defines:
     * agent
     *  tasks (prompt and appropriate metric. Metric can be based on the provided services e.g. service getting poses of the objects.), 
     * scene_configs defining e.g. start objects positions, 
     * scenarios (pairs of the tasks and scene_configs)
* Simulation needs to meet some assumptions defined by the specific EngineConnector - e.g. specific services (e.g. used for spawning objects getting objects names and poses, getting available objects to spawn); for O3DEngineConnector simulation provided as binary
* SceneConfig is a model that matches specific EngineConnector - e.g. when O3DEngineConnector has a possibility to spawn objects, O3DESceneConfig has field defining objects names and their initial poses.

boczekbartek:
@MagdalenaKotynia @maciejmajek Can we close this issue?

MagdalenaKotynia:
#436 addressing this issue has been merged to development, so I am closing the issue

PR #415, opened 2025-02-10
feat: rai_sim

MagdalenaKotynia:
## Purpose
To implement package for running simulations with different setups.
* SimulationBridge - the interface to connect with specific engine. It is responsible for scene setup, spawning, despawning objects, etc. 
* `def setup_scene` logic is based on the provided simulation_config of type specific for given bridge. There is a common configuration part that should be provided for each bridge, it is implemented as SimulationConfig class.
* SceneState is going to store the initial scene setup. 

* O3DExROS2Bridge - implementation of the SimulationBridge for O3DE with ROS2

## Issues
https://github.com/RobotecAI/rai/issues/405 - This PR covers the part responsible for connection with simulation (EngineConnector - currently named SimulationBridge). 
## Testing
* [Download binary](https://robotecai.sharepoint.com/Shared%20Documents/Forms/AllItems.aspx?id=%2FShared%20Documents%2FRAIManipulationDemoBenchmark%5Fjammyhumble%2Ezip&parent=%2FShared%20Documents&p=true&ga=1)
* Save the following content as (adjust binary_path if needed):
   - base_config1.yaml:
```
entities:
  - name: apple1
    prefab_name: apple
    pose:
      translation:
        x: 0.5
        y: 0.0
        z: 0.05
      rotation:
        x: 0.0
        y: 0.0
        z: 0.0
        w: 1.0
  - name: cube1
    prefab_name: green_cube
    pose:
      translation:
        x: 0.5
        y: 0.3
        z: 0.05
      rotation:
        x: 0.0
        y: 0.0
        z: 0.0
        w: 1.0
```

   - base_config2.yaml:
   
```
entities:
  - name: object1
    prefab_name: carrot
    pose:
      translation:
        x: 0.5
        y: -0.3
        z: 0.05
      rotation:
        x: 0.0
        y: 0.0
        z: 0.0
        w: 1.0
  - name: object2
    prefab_name: corn
    pose:
      translation:
        x: 0.5
        y: 0.3
        z: 0.05
      rotation:
        x: 0.0
        y: 0.0
        z: 0.0
        w: 1.0
```

   - o3de_config.yaml
     
```
binary_path: RAIManipulationDemoBenchmark_jammyhumble/RAIManipulationDemo/RAIManipulationDemo.GameLauncher
robotic_stack_command: ros2 launch examples/manipulation-demo-no-binary.launch.py 
```

* Run the following python script (adjust paths to yaml configs if needed):
```
import rclpy
from rai_sim.o3de.o3de_bridge import O3DExROS2Bridge, O3DExROS2SimulationConfig
from rai.communication.ros2.connectors import ROS2ARIConnector
import time
from pathlib import Path

if __name__ == "__main__":

    try:
        rclpy.init()
        connector = ROS2ARIConnector()
        o3de = O3DExROS2Bridge(connector)

        scene_config  = O3DExROS2SimulationConfig.load_config(base_config_path=Path("base_config1.yaml"), bridge_config_path=Path("o3de_config.yaml"))
        o3de.setup_scene(scene_config)
        time.sleep(5)

        scene_config  = O3DExROS2SimulationConfig.load_config(base_config_path=Path("base_config2.yaml"), bridge_config_path=Path("o3de_config.yaml"))
        o3de.setup_scene(scene_config)

        time.sleep(5)

        o3de.shutdown()
        
        
        connector.shutdown()
        rclpy.shutdown()

    except Exception as e:
        raise e
    finally:
        o3de.shutdown()
        
        
        connector.shutdown()
        rclpy.shutdown()
```

## Summary by CodeRabbit

- **New Features**
  - Introduced a new simulation module that enhances scene configuration and entity management for a more seamless simulation experience.
  - Added improved integration with a simulation engine for robust process control and dynamic entity handling.

- **Documentation**
  - Updated package metadata to include clear licensing, versioning, and dependency requirements.


src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,174 @@
...
+            logger.debug(f"Sending SIGINT to process {self.current_process.pid}")
+            self.current_process.send_signal(signal.SIGINT)
+
+            parent = psutil.Process(self.current_process.pid)
+            children = parent.children(recursive=True)
+            processes = children + [parent]
+
+            _, alive = psutil.wait_procs(processes, timeout=3)
+
+            # NOTE (mkotynia) kill ros2
```
maciejmajek: ros2 should never be killed
we can kill the robotic stack when sim has finished

adamdbrw: Call rclpy shutdown or similar, please do it standard way not brute force hacks

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,174 @@
...
+            except Exception as e:
+                logger.debug(f"Failed to kill process group: {e}")
+
+        finally:
+            self.current_process = None
+
+    def __del__(self):
+        self.shutdown()
+
+    def get_available_spawnable_names(self) -> list[str]:
```
maciejmajek: Please add a try except for n_of possible retries. Same applies to other methods.

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,174 @@
...
+            _, alive = psutil.wait_procs(processes, timeout=3)
+
+            # NOTE (mkotynia) kill ros2
+            for process in alive:
+                logger.debug(f"Force killing process {process.pid}, {process.name()}")
+                try:
+                    process.kill()
+                except Exception as e:
+                    logger.warning(f"Failed to kill process {process.pid}: {e}")
+                    continue
```
maciejmajek: Is the continue needed here?

src/rai_sim/rai_sim/engine_connector.py
```diff
@@ -0,0 +1,97 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List, Optional
+
+import yaml
+from geometry_msgs.msg import Point, Pose, Quaternion
```
boczekbartek: I wouldn't rely on ros2 messages on this abstraction level. 
As I understand this is a generic interface to connect to simulation engines.
Some simulation engines have a build-in ros2 support (o3de, gazebo, isaacsim through bridge), but some don't (Mujoco, Genesis).

src/rai_sim/rai_sim/engine_connector.py
```diff
@@ -0,0 +1,97 @@
...
+
+class Entity(BaseModel):
+    name: str
+    prefab_name: str
+    pose: Any
+    # TODO (mkotynia) consider whether create base model for poses instead of using Any as a workaround
+
+    @field_validator("pose", mode="after")
+    @classmethod
+    def convert_to_pose(cls, value: Dict[str, Any]) -> Pose:
```
boczekbartek: I think this workaround is not needed if we skip using `geometry_msgs.Pose`

src/rai_sim/rai_sim/engine_connector.py
```diff
@@ -0,0 +1,97 @@
...
+            orientation=Quaternion(**rotation) if rotation else Quaternion(),
+        )
+
+
+class SceneConfig(BaseModel):
+    """
+    Setup of scene - arrangmenet of objects, interactions, environment etc.
+    """
+
+    binary_path: Optional[str]
```
boczekbartek: `binary_path` seem to be restrictive. How about allowing a command or script. 
Looking at the implementation it seems like the rest of the code should handle it, but variable name suggests that only a binary file is supported

src/rai_sim/rai_sim/engine_connector.py
```diff
@@ -0,0 +1,97 @@
...
+        translation = value.get("translation", {})
+        rotation = value.get("rotation", {})
+
+        return Pose(
+            position=Point(**translation) if translation else Point(),
+            orientation=Quaternion(**rotation) if rotation else Quaternion(),
+        )
+
+
+class SceneConfig(BaseModel):
```
boczekbartek: `SceneConfig` contains a information on how to run a simulator. Shouldn't it be named `SimulationConfig`

src/rai_sim/rai_sim/engine_connector.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    Setup of scene - arrangmenet of objects, interactions, environment etc.
```
boczekbartek: I don't know if I understand the purpose of adding both `SceneSetup` and `SceneConfig`. This is what I assume based on the implementation:
- `SceneConfig` contains initial objects arrangement
- `SceneSetup` contains current objects arrangement

For example in a basic benchmark as a part of task (so initial state) we would have SceneConfig and some language instruction.
Then we run the rai agent and check if `SceneSetup` is as expected. If it's fine, then benchmark is passed.

I am not sure what are `interactions`.

If my understanding is correct I propose to:
- Rename `SceneConfig` to `SimulationConfig`
- Adjust docs - `Setup of scene - set initial arrangement of objects in the environment`
    - be aware of typos
- Rename `SceneSetup` to `SceneState`

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,174 @@
...
+                self._despawn_entity_by_id(self.entity_ids[entity])
+        self.entity_ids = {}
+        time.sleep(3)
+        for entity in scene_config.entities:
+            self._spawn_entity(entity)
+        # TODO (mkotynia) handle SceneSetup
+        return SceneSetup(entities=scene_config.entities)
+
+    def launch_binary(self, binary_path: str):
+        # NOTE (mkotynia) ros2 launch command with binary path, to be refactored
```
boczekbartek: Connected to my previous comments

src/rai_sim/rai_sim/engine_connector.py
```diff
@@ -0,0 +1,97 @@
...
+    Load the scene configuration from a YAML file.
+    """
+    try:
+        with open(file_path, "r") as file:
+            content = yaml.safe_load(file)
+
+        return SceneConfig(**content)
+
+    except Exception as e:
+        raise e
```
boczekbartek: I think this code has no effect. It catches exceptions and then raises them again

src/rai_sim/rai_sim/engine_connector.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+
+    binary_path: Optional[str]
+    entities: List[Entity]
+
+
+class SceneSetup(BaseModel):
+    """
+    Info about entities in the scene (positions, collisions, etc.)
+    """
```
boczekbartek: Consider renaming to `SceneState` (more info in previous comment). 
As for collisions, I think they are not included in the current implementation. How about adding a NOTE that they are meant to be added here?

boczekbartek (changes requested):
@MagdalenaKotynia Thank you for this PR! 

- I left some comments to the code.
- Also I was able to run the example from the PR description, but it didn't work in the first run - I think the binary was loading too long. See my error below. Do you check if the binary is ready to spawn objects?
- it would be good to add an info to the description of the PR that example will only work if you setup [rai-manipulation-demo](https://github.com/RobotecAI/rai/blob/main/docs/demos/manipulation.md) first (because for e.g. [this line](https://github.com/RobotecAI/rai/blob/664adfad93e032ed03dc7de61aad553d7783cdf9/examples/manipulation-demo.launch.py#L67))
```
(rai-framework-py3.10) robo-pc-005 ➜  rai git:(feat/benchmarking) ✗ python run.py                      
/home/bboczek/.cache/pypoetry/virtualenvs/rai-framework-2NHnaPvx-py3.10/lib/python3.10/site-packages/pydub/utils.py:170: RuntimeWarning: Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work
  warn("Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work", RuntimeWarning)
2025-02-12 16:21:02 robo-pc-005 rai_sim.o3de.o3de_connector[159409] INFO Running command: ['ros2', 'launch', 'examples/manipulation-demo.launch.py', 'game_launcher:=/home/bboczek/Downloads/RAIManipulationDemo/RAIManipulationDemo.GameLauncher']
Traceback (most recent call last):
  File "/u24/home/bboczek/projects_u24/01_internal/01_repos/rai/run.py", line 16, in <module>
    o3de.setup_scene(scene_config)
  File "/u24/home/bboczek/projects_u24/01_internal/01_repos/rai/src/rai_sim/rai_sim/o3de/o3de_connector.py", line 161, in setup_scene
    self._spawn_entity(entity)
  File "/u24/home/bboczek/projects_u24/01_internal/01_repos/rai/src/rai_sim/rai_sim/o3de/o3de_connector.py", line 90, in _spawn_entity
    entity.pose, self.connector.get_transform("odom", "world")
  File "/u24/home/bboczek/projects_u24/01_internal/01_repos/rai/src/rai_core/rai/communication/ros2/connectors.py", line 176, in get_transform
    raise LookupException(
tf2.LookupException: Could not find transform from world to odom in 5.0 seconds
Exception in thread Thread-1 (spin):
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 294, in spin
    self.spin_once()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 794, in spin_once
    self._spin_once_impl(timeout_sec)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 786, in _spin_once_impl
    self._executor.submit(handler)
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 169, in submit
    raise RuntimeError('cannot schedule new futures after '
RuntimeError: cannot schedule new futures after interpreter shutdown
Exception ignored in: <function O3DEngineConnector.__del__ at 0x7e176e350790>
Traceback (most recent call last):
  File "/u24/home/bboczek/projects_u24/01_internal/01_repos/rai/src/rai_sim/rai_sim/o3de/o3de_connector.py", line 75, in __del__
  File "/u24/home/bboczek/projects_u24/01_internal/01_repos/rai/src/rai_sim/rai_sim/o3de/o3de_connector.py", line 64, in shutdown
AttributeError: 'NoneType' object has no attribute 'warning'
```

src/rai_sim/rai_sim/engine_connector.py
maciejmajek: Correct, the pose/orientation should not depend on ros2.

maciejmajek:
@coderabbitai full review

src/rai_sim/rai_sim/engine_connector.py
pijaro: Consider using `pathlib` to store paths.

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,205 @@
...
+            return
+        parent = psutil.Process(self.current_process.pid)
+        children = parent.children(recursive=True)
+
+        # NOTE (mkotynia) terminating binary
+        for child in children:
+            logger.debug(f"Terminating child process {child.pid}, {child.name()}")
+            child.terminate()
+
+        _, alive = psutil.wait_procs(children, timeout=3)
```
pijaro: Increase the timeout, some applications might take up considerable amount of time to clean itself and shutdown.

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,205 @@
...
+            w=pose.orientation.w,  # type: ignore
+        )
+
+        pose_model = PoseModel(translation=translation, rotation=rotation)
+
+        return pose_model
+
+    def launch_binary(self, binary_path: str):
+        # NOTE (mkotynia) ros2 launch command with binary path, to be refactored
+        command = shlex.split(binary_path)
```
pijaro: This looks like you allow the `binary_path` to have the start arguments as well.

Is it safe? Shouldn't we treat binary path and arguments separately?

`binary_path` should be just a path. Otherwise, more appropriate name should be used, like `command` or something similar.

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,205 @@
...
+            if scene_config.binary_path:
+                self.launch_binary(scene_config.binary_path)
+            else:
+                raise Exception("No binary path provided")
+            self.current_binary_path = scene_config.binary_path
+        else:
+            for entity in self.entity_ids:
+                self._despawn_entity_by_id(self.entity_ids[entity])
+        self.entity_ids = {}
+        time.sleep(3)
```
pijaro: Hard coded sleep looks so wrong. We should be able to detect somehow that the binary has started and is ready to receive commands

I can guarantee that some sims will need more than N seconds to start.

src/rai_sim/README.md
pijaro, line 1: Please fill the readme. At least some core information.

src/rai_sim/rai_sim/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2024 Robotec.AI
```
pijaro: Uber nit: shouldn't we change the date to 2025? :)

src/rai_sim/rai_sim/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""RAI Simulations package."""
+
+__version__ = "0.0.1"
```
pijaro: Just a thought: can we grab this from `.toml` file? Or should we keep track version number in two different places?

src/rai_sim/rai_sim/engine_connector.py
```diff
@@ -0,0 +1,123 @@
...
+    entities: List[Entity]
+
+
+class EngineConnector(ABC):
+    """
+    Responsible for communication with simulation.
+    """
+
+    def __init__(self):
+        pass
```
pijaro: There are no shared attributes across the connectors, therefore this can be removed.
```suggestion
```

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,205 @@
...
+    EngineConnector,
+    Entity,
+    PoseModel,
+    Rotation,
+    SceneConfig,
+    SceneSetup,
+    Translation,
+)
+
+logger = logging.getLogger(__name__)
```
pijaro: Is this consistent across all rai packages? Looks a bit too barebone to me.

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,205 @@
...
+        for child in children:
+            logger.debug(f"Terminating child process {child.pid}, {child.name()}")
+            child.terminate()
+
+        _, alive = psutil.wait_procs(children, timeout=3)
+        if alive:
+            for child in alive:
+                logger.warning(f"Force killing child process PID: {child.pid}")
+                child.kill()
+        # NOTE (mkotynia) terminating ros2 launch
```
pijaro: This is engine connector, why do you mention `ros2 launch`? what do `ros2 launch` have to do with `O3DEngineConnector`?

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,205 @@
...
+                    "z": pose.orientation.z,  # type: ignore
+                    "w": pose.orientation.w,  # type: ignore
+                },
+            },
+        }
+
+        msg = ROS2ARIMessage(payload=msg_content)
+
+        response = self.connector.service_call(
+            msg, target="spawn_entity", msg_type="gazebo_msgs/srv/SpawnEntity"
```
pijaro: I'm not feeling coupling the ROS 2 interface with O3DEngineConnector. Semantically, talking about "o3de connector" - shouldn't we care about connecting rai with o3de only (o3de interfaces, launch arguments, parameters, configuration)?

We assumed here that the O3DE comes with ROS 2 interface and the simulation follows certain standard/implementation. 

What if someone have different API to control O3DE simulation?

Maybe we should make a distinguish between engine, and provided simulation interface?

If we bring a concrete implementation of the "connector" (in this case, an O3DE based sim that follows a standard and uses ROS 2 gem), then we should consider changing a name to something more concrete.

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def launch_binary(self, binary_path: str):
+        # NOTE (mkotynia) ros2 launch command with binary path, to be refactored
+        command = shlex.split(binary_path)
+        logger.debug(f"Running command: {command}")
+        self.current_process = subprocess.Popen(
+            command,
+        )
+        if not self.has_process_started():
+            raise RuntimeError("Process did not start in time.")
```
pijaro: Inconsistent use of exception types. `setup_scene` throws an `Exception`, but here we are using `RuntimeError`. Both are "on the same level":

```python
if scene_config.binary_path:
    self.launch_binary(scene_config.binary_path) # can raise RuntimeError
else:
    raise Exception("No binary path provided")
```

src/rai_sim/rai_sim/o3de/o3de_connector.py
maciejmajek: > we should consider changing a name to something more concrete

Right. We have discussed this topic outside of git, our O3DE + ROS2 combo relies on a specific O3DE setup. @knicked is currently working on a o3de requirements document.

src/rai_sim/rai_sim/__init__.py
maciejmajek: +1, track with pyproject.toml only

MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/1ca120366d17d647338bd6dc172d1963e60a61b6

src/rai_sim/rai_sim/__init__.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/ac7fd2147fde927ad8606793d3369096d8e740c2

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/2b4286ec67fedd58955f727f34a0f66bccfdb583#diff-dc20158609deed61acd0c22c03e989b7356f79311a7be8dc9227639b9904ca16R59

src/rai_sim/rai_sim/engine_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/db35adbe1842218e8298e4bca6e86cb5bef17c91 and  https://github.com/RobotecAI/rai/pull/415/commits/0c3ea02d1b5781463ea291315033e959edb503c6

src/rai_sim/rai_sim/engine_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/db35adbe1842218e8298e4bca6e86cb5bef17c91 and  https://github.com/RobotecAI/rai/pull/415/commits/0c3ea02d1b5781463ea291315033e959edb503c6

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/2b4286ec67fedd58955f727f34a0f66bccfdb583

src/rai_sim/rai_sim/engine_connector.py
MagdalenaKotynia: true, removed in: https://github.com/RobotecAI/rai/pull/415/commits/d09cbbd773ba1fc46c3a89e8422b37eb1485fc31

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: @maciejmajek do you have some convention about logging handling? I see that some classes have their logger attribute but some don't.

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: I added the check if scene_config.binary_path outside the launch_binary method because it seemed for me to be a clearer solution than allowing the call of the launch_binary method with an optional binary_path argument and checking it inside this method. Currently the problem is outdated because after yesterday discussion with @maciejmajek we decided to separate common config attributes to base SimulationConfig class and to define other attributes in custom classes for each connector (implemented in https://github.com/RobotecAI/rai/pull/415/commits/ab5bf1fc5ed81331dabdb8d0e484f6d9e294ce43). As a result, the binary_path is now mandatory for this connector and there is no need to check it inside the connector (it is handled when loading config content to pydantic model).

src/rai_sim/rai_sim/engine_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/8ed84b30b71b56eb3efcc44d4069461b151e6140

src/rai_sim/rai_sim/engine_connector.py
MagdalenaKotynia: Yes, your understanding is correct. I applied the suggested renaming changes as they better reflect these classes' meaning. 
Interactions and environment were here examples of additional parameters that could be configured in the future. As it seems to be misleading, I removed it from the docs and left what is currently possible to be configured and added a note that it can be extended.

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: I renamed EngineConnector with SimulationConnector and O3DEngineConnector to O3DExROS2Connector to better reflect the meaning - that it is O3DE with ROS2 interface.

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: I removed this hard-coded sleep, I added a check if the process has started inside launch_binary/launch_robotic_stack defs. I am wondering whether this check is enough to be sure that the binary has been launched. Maybe there should be some additional check whether some topics started to send messages, e.g. clock, what do you think?

src/rai_sim/rai_sim/engine_connector.py
MagdalenaKotynia: Removed it.

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: I separated launching binary and launching robotic stack in https://github.com/RobotecAI/rai/pull/415/commits/21b31a448dab7690f09a6e49c12234c1033c3c29

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: it is related to this discussion https://github.com/RobotecAI/rai/pull/415#discussion_r1954446498. Currently, it is a simulation connector.

src/rai_sim/README.md
MagdalenaKotynia: Done in https://github.com/RobotecAI/rai/pull/415/commits/13bf3582a6c2d3f136355c025b99b83ac4accc7e

src/rai_sim/rai_sim/config_loader.py
```diff
@@ -0,0 +1,50 @@
...
+from pathlib import Path
+from typing import Any, Set, Type
+
+import yaml
+
+from rai_sim.o3de.o3de_connector import O3DExROS2SimulationConfig
+from rai_sim.simulation_connector import SimulationConfig, SimulationConfigT
+
+CONFIG_REGISTRY: Set[Type[SimulationConfig]] = {O3DExROS2SimulationConfig}
+
```
rachwalk: This part makes it so that for a user defining new configs changing the package source code is required. I'd reccomend using `isinstance` instead.

src/rai_sim/rai_sim/config_loader.py
```diff
@@ -0,0 +1,50 @@
...
+
+def load_base_config(path: Path) -> SimulationConfig:
+    """Load the base configuration file"""
+    with open(path) as f:
+        content = yaml.safe_load(f)
+    return SimulationConfig(**content)
+
+
+def load_simulation_config(
```
rachwalk: I think it would be more maintainable if instead of a factory function like this, SimulationConfig had a classmethod constructor. Refer to: https://stackoverflow.com/questions/61789002/pydantic-and-constructors

src/rai_sim/rai_sim/o3de/o3de_connector.py
rachwalk: A class should take a logger as an optional argument (validation for both ROS and python loggers, and if None set it to default logger.

src/rai_sim/rai_sim/o3de/o3de_connector.py
rachwalk: I see that you've implemented the retrial mechanism using timing. It is rather non standard, and can lead to issues with reproducability. Stanard practice is using a set amount of retries (typically defaults to 3)

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,232 @@
...
+                    "z": pose.orientation.z,  # type: ignore
+                    "w": pose.orientation.w,  # type: ignore
+                },
+            },
+        }
+
+        msg = ROS2ARIMessage(payload=msg_content)
+
+        response = self._try_service_call(
+            msg, target="spawn_entity", msg_type="gazebo_msgs/srv/SpawnEntity"
```
rachwalk: Is it possible for this service to return a value, while failing to spawn the entity? Or does getting a response guarantee success on the spawning task?

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,232 @@
...
+
+        response = self._try_service_call(
+            msg, target="delete_entity", msg_type="gazebo_msgs/srv/DeleteEntity"
+        )
+        if response:
+            self.spawned_entities.remove(entity)
+
+    def get_object_pose(self, entity: SpawnedEntity) -> PoseModel:
+        object_name = entity.name
+        object_frame = object_name + "/"
```
rachwalk:
```suggestion
        object_frame = entity.name + "/"
```

src/rai_sim/rai_sim/o3de/o3de_connector.py
```diff
@@ -0,0 +1,232 @@
...
+        command = shlex.split(robotic_stack_command)
+        logger.info(f"Running command: {command}")
+        self.current_robotic_stack_process = subprocess.Popen(
+            command,
+        )
+        if not self._has_process_started(self.current_robotic_stack_process):
+            raise RuntimeError("Process did not start in time.")
+
+    def _has_process_started(
+        self, process: subprocess.Popen[Any] | None, timeout: int = 15
```
rachwalk: Why is `None` a valid argument for this method? If `process=None` this method is equivalent to `sleep(15)`, so even if it's required to handle such an argument it would be warranted to early quit instead of waiting for 15 seconds.

src/rai_sim/rai_sim/simulation_connector.py
```diff
@@ -0,0 +1,123 @@
...
+    pose: PoseModel
+
+
+class SpawnedEntity(Entity):
+    id: str
+
+
+class SimulationConfig(BaseModel):
+    """
+    Setup of simulation - arrangemenet of objects in the environment. # NOTE (mkotynia) can be extended by other attributes
```
rachwalk: Move the comment outside of docstring - if you do it like this, the comment is interpreted as part of the docstring and shows up when looking it up.

src/rai_sim/rai_sim/simulation_connector.py
```diff
@@ -0,0 +1,123 @@
...
+    def check_unique_names(cls, entities: List[Entity]) -> List[Entity]:
+        names = [entity.name for entity in entities]
+        if len(names) != len(set(names)):
+            raise ValueError("Each entity must have a unique name.")
+        return entities
+
+
+class SceneState(BaseModel):
+    """
+    Info about current entities' state in the scene # NOTE (mkotynia) can be extended by other attributes
```
rachwalk: ditto

src/rai_sim/rai_sim/utils.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+    rotation = Rotation(
+        x=pose.orientation.x,  # type: ignore
+        y=pose.orientation.y,  # type: ignore
+        z=pose.orientation.z,  # type: ignore
+        w=pose.orientation.w,  # type: ignore
+    )
+
+    return PoseModel(translation=translation, rotation=rotation)
```
rachwalk: Consider implementing this as a class method of PoseModel instead. Having an `utils.py` to have a something which is essentially a constructor is not the best practice.

rachwalk (changes requested):
Thank you for this PR in general it looks good. I left a few comments regarding coding practices, I will continue the review once these are implemented.

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: TODO: handle if spawn fails

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: TODO: check if process is not None outside _has_process_started

src/rai_sim/rai_sim/config_loader.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/0f0de21cb77675c08c492932c79f573a48bf26a7

src/rai_sim/rai_sim/config_loader.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/5589c18ef3e629d208d1317ecd9d09b7135227b7

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/2a97f9ff67d6755c86ed5b67145d814c3d4ff5ef

src/rai_sim/rai_sim/simulation_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/db615de4e25f497d2afe7feb93e6b51553edba73

src/rai_sim/rai_sim/simulation_connector.py
MagdalenaKotynia: ditto

src/rai_sim/rai_sim/utils.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/04aa67b5a54541ce38133148233c68059d2e58e1

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/6f05cf796945f77ff9321e4194744d394d1a00c9

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: applied in https://github.com/RobotecAI/rai/pull/415/commits/f439bde710f85d86319538a8b85f75b8e040ff4b

src/rai_sim/rai_sim/o3de/o3de_connector.py
MagdalenaKotynia: I added errors handling in https://github.com/RobotecAI/rai/pull/415/commits/2a97f9ff67d6755c86ed5b67145d814c3d4ff5ef.
Services used in this task may return value with success == False (e.g. when not existing prefab name is given) or may raise an error (e.g. when the service with given name is not running or does not exist).

src/rai_sim/rai_sim/engine_connector.py
boczekbartek: I can see it has been renamed, resolving

src/rai_sim/rai_sim/o3de/o3de_bridge.py
```diff
@@ -0,0 +1,306 @@
...
+    SceneState,
+    SimulationBridge,
+    SimulationConfig,
+    SpawnedEntity,
+    Translation,
+)
+
+
+class O3DExROS2SimulationConfig(SimulationConfig):
+    binary_path: Path
```
boczekbartek, line 41: @MagdalenaKotynia How about [this comment](https://github.com/RobotecAI/rai/pull/415#discussion_r1952833977)?

src/rai_sim/rai_sim/engine_connector.py
MagdalenaKotynia: Currently, base SimulationConfig has only the common part for all simulations so it does not have binary_path parameter. When implementing the SimulationBridge interface, it is needed to create a custom config and define parameters for running/setup of the simulation. Therefore, binary_path parameter was moved to O3DExROS2Bridge.

src/rai_sim/rai_sim/o3de/o3de_bridge.py
MagdalenaKotynia: I've already replied https://github.com/RobotecAI/rai/pull/415#discussion_r1969518478, sorry, I missed this comment.

PR #426, opened 2025-02-19
feat: O3DE test benchmark

jmatejcz:
## Purpose
Add Test Benchmark for O3DE simulation
Add framework for easier benchmark creation

## Proposed Changes
Majority of file changes are only formatting changes, main changes are in src/rai_bench package and src/rai_sim 

This PR contains new package rai_bench. It introduces classes like  Task, Scenario and Benchmark(src/rai_bench/rai_bench/benchmark_model.py), which are frame for creating benchmarks and it also contains O3DE Test Benchmark which is a test implementation of a benchmark for O3DE engine simulation.

O3DE Test Benchmark (`src/rai_bench/rai_bench/o3de_test_bench/`), contains 2 Tasks(`tasks/`) - GrabCarrotTask and PlaceCubesTask  (these tasks implement calculating scores) and 4 scenes(`configs/`) for O3DE robotic arm simulation. 

Example of how can scenarios and benchmark be defined and run  can be found in `rai_bench/rai_bench/main.py`

Additional changes:
- Subclass of Bridge for Arm Manipulation-> O3DEngineArmManipulationBridge
- validation if required services are available -> _is_robotic_stack_ready() function in SimulationBridge
- fix a bug when get_transform returned same values every time-> https://github.com/RobotecAI/rai/issues/430

## Testing
the binary to simulation can be downloaded from here: [humble](https://robotecai.sharepoint.com/Shared%20Documents/Forms/AllItems.aspx?id=%2FShared%20Documents%2FRAIManipulationDemoBenchmark%5Fjammyhumble%20%28old%29%2Ezip&parent=%2FShared%20Documents&p=true&ga=1)

To run this benchmark you need o3de bridge config file in `src/rai_bench/rai_bench/o3de_test_bench/configs/o3de_config.yaml`, this config file should include fields:
```yaml
binary_path: ...
robotic_stack_command: ros2 launch examples/manipulation-demo-no-binary.launch.py
required_services:
  - /grounding_dino_classify
  - /grounded_sam_segment
  - /manipulator_move_to
  - /spawn_entity
  - /delete_entity
required_topics:
  - /color_image5
  - /depth_image5
  - /color_camera_info5
required_actions: []
```
to install and run:
```
poetry install --with openset
colcon build --symlink-install
source setup_shell.sh
python src/rai_bench/rai_bench/main.py
```
Logs about the progess of benchmark can be found in 'src/rai_bench/o3de_test_bench/benchmark_agent.log'

You should be able to see: 
- properly started simulation
- 4 defined scenarios
- logs about the actions taken by agent and scores achieved in each scenario
After finishing all scenarios, simulation should close and all nodes and services should also shutdown after no more than 30 seconds


imgui.ini
```diff
@@ -0,0 +1,5 @@
+[Window][Debug##Default]
+Pos=60,60
+Size=400,400
+Collapsed=0
+
```
rachwalk: Why is this file included?

src/rai_bench/README.md
rachwalk, line 1: Why is an empty README included

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,195 @@
...
+from rai.messages import HumanMultimodalMessage
+from rai_sim.simulation_bridge import (
+    SimulationBridge,
+    SimulationConfig,
+    PoseModel,
+    SpawnedEntity,
+)
+
+
+SimulationConnectorT = TypeVar("SimulationConnectorT", bound=SimulationBridge)
```
rachwalk: Rename this to avoid confusion with rai BaseConnector derived types

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,195 @@
...
+                        last_msg = msg.content
+                else:
+                    raise ValueError(f"Unexpected type of message: {type(msg)}")
+
+                self._logger.debug(f"{graph_node_name}: {last_msg}")
+                self._logger.info(f"AI Message: {msg}")
+                # TODO (jm) figure out how to get number of tool calls
+            te = time.perf_counter()
+
+            self.engine_connector.get_scene_state()
```
rachwalk: What is this for?

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+class Scenario(BaseModel):
+    """Single instances are run separatly by benchmark"""
+
+    task: Task
+    scene_config: SimulationConfig
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True
+    )  # pydantic does not support ABC classes
```
rachwalk: As far as I can see this parameter is never used, what should it be used for?

src/rai_bench/o3de_test_bench/main.py
```diff
@@ -0,0 +1,302 @@
...
+    O3DExROS2SimulationConfig,
+)
+
+from rai_interfaces.srv import ManipulatorMoveTo
+from geometry_msgs.msg import PoseStamped
+from std_msgs.msg import Header
+from pathlib import Path
+
+
+class GrabCarrotTask(Task):
```
rachwalk: It would be a good idea to create a module with Tasks so they can be imported into different benchmark runtimes.

src/rai_bench/o3de_test_bench/scene1.yaml
rachwalk, line 1: It would be a good idea to move these to a separate folder to keep things organised.

src/rai_bench/o3de_test_bench/o3de_config.yaml
rachwalk, line 1: Why is this file included?

src/rai_bench/o3de_test_bench/main.py
```diff
@@ -0,0 +1,302 @@
...
+            scene_config=red_cubes_scene_config,
+        ),
+        Scenario(
+            task=RedCubesTask(logger=bench_logger), scene_config=red_cubes_scene_config
+        ),
+        Scenario(
+            task=RedCubesTask(logger=bench_logger),
+            scene_config=multiple_cubes_scene_config,
+        ),
+    ]
```
rachwalk: It would be a good idea to have these load dynamically to allow for easy configuration of scenes/tasks.

rachwalk (changes requested):
Overall good job, I have left a few comments regarding code structure, and I am unsure if all the committed files should be included.

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,195 @@
...
+SimulationConnectorT = TypeVar("SimulationConnectorT", bound=SimulationBridge)
+loggers_type = Union[RcutilsLogger, logging.Logger]
+
+
+class EntitiesMismatchException(Exception):
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
+
+
+class Task(ABC, Generic[SimulationConnectorT]):
```
MagdalenaKotynia: Task class should not be bound with any specific SimulationBridge, it should be reused with different implementations of SimulationBridge. Please apply appropriate typing for this class and the calculate_result method.

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,195 @@
...
+        logger: loggers_type | None = None,
+    ) -> None:
+        self.engine_connector: SimulationBridge
+        self.tasks: list[Task] = []
+        self.scenarios = enumerate(iter(scenarios))
+        self.results = []
+        if logger:
+            self._logger = logger
+        else:
+            self._logger = logging.getLogger(__name__)
```
MagdalenaKotynia: I think that class Benchmark should be redefined. Currently, it allows for benchmarking scenarios and agents but it does not allow for benchmarking simulations using different implementations of SimulationBridge interface (see requirements in https://github.com/RobotecAI/rai/issues/405#issue-2835184006). What about creating class Benchmark as an interface for implementing specific benchmarks, e.g. SimBenchmark to benchmark different simulations with locked llm model and agent structure; or AgentBenchmark to benchmark different Agents with locked simulation, llm model; etc?

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,195 @@
...
+    Defined by a set of scenarios to be done
+    """
+
+    def __init__(
+        self,
+        scenarios: list[Scenario],
+        logger: loggers_type | None = None,
+    ) -> None:
+        self.engine_connector: SimulationBridge
+        self.tasks: list[Task] = []
```
MagdalenaKotynia: For what is self.tasks? It is not used, tasks are already included in the scenarios.

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,195 @@
...
+                else:
+                    raise ValueError(f"Unexpected type of message: {type(msg)}")
+
+                self._logger.debug(f"{graph_node_name}: {last_msg}")
+                self._logger.info(f"AI Message: {msg}")
+                # TODO (jm) figure out how to get number of tool calls
+            te = time.perf_counter()
+
+            self.engine_connector.get_scene_state()
+            result = scenario.task.calculate_result(self.engine_connector)
```
MagdalenaKotynia: Please append result to the list of results. I would also consider a different structure for storing results, to also store information about the experiment to which a given result belongs.

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,195 @@
...
+        self.engine_connector: SimulationBridge
+        self.tasks: list[Task] = []
+        self.scenarios = enumerate(iter(scenarios))
+        self.results = []
+        if logger:
+            self._logger = logger
+        else:
+            self._logger = logging.getLogger(__name__)
+
+    def run_next(self, agent):
```
MagdalenaKotynia: I think it would be useful to calculate the result also at the beginning of each scenario, to be able to recognize if a good result at the end is not a coincidental effect of a favorable initial arrangement of elements on the scene.
I see that it is somehow handled in the implementations of calculate_result for each task, but I think calculating results with the same metric at the beginning and at the end of the scenario would be a simpler approach - the user wouldn't need to think about it when writing metrics.

src/rai_bench/o3de_test_bench/main.py
```diff
@@ -0,0 +1,302 @@
...
+            self.logger.info("No objects to manipulate, returning score 1.0")
+            return 1
+        else:
+            self.logger.info(
+                f"corrected_objects: {corrected_objects}, misplaced_objects: {misplaced_objects}, unchanged_correct: {unchanged_correct}, displaced_objects: {displaced_objects}"
+            )
+            return (corrected_objects + unchanged_correct) / num_of_objects
+
+
+def request_to_base_position() -> ManipulatorMoveTo.Request:
```
MagdalenaKotynia: What about moving this function to O3DEngineArmManipulationBridge and parametrizing it with the initial arm pose?

src/rai_bench/o3de_test_bench/main.py
```diff
@@ -0,0 +1,302 @@
...
+            self.logger.info("No objects to manipulate, returning 1.0")
+            return 1.0
+        else:
+            self.logger.info(
+                f"corrected_objects: {corrected_objects}, misplaced_objects: {misplaced_objects}, unchanged_correct: {unchanged_correct}, displaced_objects: {displaced_objects}"
+            )
+            return (corrected_objects + unchanged_correct) / num_of_objects
+
+
+class RedCubesTask(Task):
```
MagdalenaKotynia: Rename the class accordingly to the prompt of the task.

src/rai_bench/o3de_test_bench/main.py
```diff
@@ -0,0 +1,302 @@
...
+            scene_state.entities, prefab_types=["carrot"]
+        )
+        num_of_objects = len(initial_carrots)
+
+        if num_of_objects != len(final_carrots):
+            raise EntitiesMismatchException(
+                "Number of initially spawned entities does not match number of entities present at the end."
+            )
+
+        for ini_carrot in initial_carrots:
```
MagdalenaKotynia: This check is strictly matching our simulation where we know how the table is located with respect to the origin. Maybe it's overkill for now but I would consider parametrizing it somehow. Maybe the calculate_result method should take some kwargs to be able to reuse the task for simulations with different parameters key for computing result (e.g. in this case the info about coordinates and size of the table top).

src/rai_bench/o3de_test_bench/main.py
```diff
@@ -0,0 +1,302 @@
...
+                            )
+                    break
+            else:
+                raise EntitiesMismatchException(
+                    f"Entity with name: {ini_carrot.name} which was present in initial scene, not found in final scene."
+                )
+
+        if num_of_objects == 0:
+            self.logger.info("No objects to manipulate, returning 1.0")
+            return 1.0
```
MagdalenaKotynia: I think we should not return 1.0 when there are no objects to manipulate. Maybe None would be better for such situations.

I would also suggest moving this check before for loop - I know that it does not have any effect but IMO it will be more logical to check this condition before going further.

src/rai_bench/o3de_test_bench/main.py
```diff
@@ -0,0 +1,302 @@
...
+
+        scene_state = engine_connector.get_scene_state()
+
+        initial_carrots = self.filter_entities_by_prefab_type(
+            engine_connector.spawned_entities, prefab_types=["carrot"]
+        )
+        final_carrots = self.filter_entities_by_prefab_type(
+            scene_state.entities, prefab_types=["carrot"]
+        )
+        num_of_objects = len(initial_carrots)
```
MagdalenaKotynia: I suggest renaming to num_initial_carrots to be clearer

MagdalenaKotynia (changes requested):
Thank you for this PR. I have left some comments. 
Apart from that:

* Please apply new naming convention for simulation connectors (see https://github.com/RobotecAI/rai/pull/415, SimulationConnector was renamed to SimulationBridge to avoid confusion with the BaseConnector and its subclasses.

* Please fix typing where possible

Ideas for the future improvement:
* I think it would be very beneficial to save output from camera (from the whole scenario or at least beginning and end of the scenario) to be able to verify metrics with human assessment. It also could be useful for presenting and reporting the results.

* When the final implementation of benchmark interface will be established, the next step could be to design and implement some structure for saving and storing the results (results + info about what was benchmarked + camera images + metadata).

jmatejcz:
> Overall good job, I have left a few comments regarding code structure, and I am unsure if all the committed files should be included.

the additiional files propably come from rebases, but you are right, they are not part of this PR, so  I will remove them

imgui.ini
jmatejcz: this file is generated by simulation i guees, i can add it to gitignore

src/rai_bench/README.md
jmatejcz: filled readme

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: model_config field is used to define additional configurations to a model. In this case the param arbitrary_types_allowed=True, allows a field with an abstract class like SimulationConfig. Without it pydantic will throw validation error

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: you are right, removing...

src/rai_bench/o3de_test_bench/main.py
jmatejcz: Do you mean to create a separate folder still inside o3de_test_bench? or to define specific tasks in rai_bench package?

src/rai_bench/o3de_test_bench/main.py
jmatejcz: do you mean loading from cmd arguments, from config file, or what?

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: Okay so i will modify the Task to take SimulationConfig as argument, and from this i will extract initial_positions of objects.

jmatejcz: Second thing is - as spawned_entities are unique to O3De , methods of SimulaitonBride should return Entities, not spawnedEntities

src/rai_bench/o3de_test_bench/main.py
jmatejcz: I don't know how the cooridinates work in other simulation engines, so it's difficult to make it work. I think for now leave it as this is, maybe in future when we will work with different engines, I will take this into cosideration

src/rai_bench/o3de_test_bench/main.py
jmatejcz: checking before the loop is a valid point. About the score, i think this should always return score. The score would be 1.0 in this situation as there is nothing to be done, but the agent still can be evaluated by number of tool calls and total time

src/rai_bench/o3de_test_bench/main.py
jmatejcz: i considered this, but there would be a lot of args, but yes it can be done

src/rai_bench/o3de_test_bench/o3de_config.yaml
jmatejcz: it is needed to simulation configuration

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: why it doesn't allow to use different engines? You can assign any Bridge you want

src/rai_bench/rai_bench/benchmark_model.py
MagdalenaKotynia: I think that Task does not need to take SimulationConfig as argument - Scenario uses SimulationConfig and bounds it with Task (btw please rename scene_config to simulation_config according to the changes in naming convention). Task itself should be SimulationConfig-agnostic, current interface is ok, just typing is inappropriate. 

Second, yes, good point. I noticed it yesterday when reviewing your PR and I decided to move spawned_entities to base SimulationBridge class because it is needed for all simulations if we want to compare them. Please rebase to these changes.

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: i will add it as arg to constructor, but the logic stays the same

MagdalenaKotynia: Yes, I  can use any Bridge I want but I cannot benchmark bridges - I cannot pass more than one bridge and test them with locked other parameters.

jmatejcz:
add new occuring issue to the main comment of this PR

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: I get your point, you pobably refering to the special case where there is nothing to move and cases where the inital setup is almost proper. But not moving objects that are already in a good place is also a good behaviour, thats why agent can achive 0.5/1 without moving anything, because it is better than displacing objects. For me, the main souce of calculation metrics are these: corrected_objects, misplaced_objects, unchanged_correct, displaced_objects. The score can be than calculated in different manners, for example punishing more displacing objects, but it is specific implementation. So to summarize i know it can be confusing that score is 1.0 when agent done nothing, but it is specific to scene and implementation of task. If user wants to calculate it like that, so be it. I can add the calculation at the start, but it seems weird to me

src/rai_bench/o3de_test_bench/main.py
MagdalenaKotynia: I think that you don't need to move this function as it is but rather refactor this solution somehow to store initial position inside the O3DEngineArmManipulationBridge and reuse it with move_arm when needed.

src/rai_bench/o3de_test_bench/main.py
MagdalenaKotynia: I didn't mean other simulation engines but e.g. another o3de robotic arm simulation where the table is located in another way with respect to origin than the current simulation we use, e.g. (x,y) = 0 is not on the  middle of the table as it is assumed in this example. I think we can leave it for now, general differentiating where is left and right is quite hard problem IMO, please just leave a NOTE with info that it is example for particular demo.

src/rai_bench/o3de_test_bench/main.py
rachwalk: I mean a functionality to auto create scenarios given a list of Tasks and a list of scene paths

src/rai_bench/o3de_test_bench/main.py
rachwalk: Yes I mean create a separate folder with __init__.py (i.e. a submodule) inside the `o3de_test_bench`

src/rai_bench/rai_bench/benchmark_model.py
rachwalk: Please mark this field clearly with a `# TODO:` explaining functioanality this field is intended for.

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: oh, okay to maybe accepting list of engines, and running all scenarios for every one of them?

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: done

src/rai_bench/o3de_test_bench/main.py
jmatejcz: okay, done

src/rai_bench/o3de_test_bench/main.py
jmatejcz: okay

src/rai_bench/rai_bench/benchmark_model.py
rachwalk: At this point there is no need for such functionality, as long as the benchmark is defined as list[scene] x list[task] x model_config, this can be separately replicated against different engines in separate runs. I would see this functionality being a part of benchmark as overkill.

src/rai_bench/rai_bench/benchmark_model.py
MagdalenaKotynia: I think that if we calculate the result at the beginning and at the end and compare them, we can also see whether the result has been worsened compared to the initial state. 

Apart from that, current implementation does not take into account all quite probable situations, e.g. 
Initial state of cubes is that two of them are adjacent at one edge of the table, and two of them are adjacent on the second edge of the table. 
**Scenario 1**: Agent does nothing. The score will be **1**, because every cube has one adjacent.
**Scenario 2**: Arm moves one cube from one edge to another and as result we have 3 adjacent cubes on one edge and one cube without any adjacent. The result will be **0.75**. The metric will classify the moved cube as **unchanged_correct** and the cube left without any adjacent as **displaced_objects** what is not true. 

In Scenario 2 agent took some action to make situation better but the result score is worse than that in Scenario 1. Therefore I would suggest to calculate the result more generally with some methods checking clusterization of the objects.
This is also why I see advantages of having the result from the beginning and from the end of the scenario. 

Uh, hard task ;)

src/rai_bench/o3de_test_bench/main.py
jmatejcz: i moved tasks to separate folder, also moved whole o3de_test_branch to rai_bench package so the benchamark/ tasks can be imported across project. main.py to setup and run benchmark moved to src/rai_bench/main.py, so still outside package

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: i refactored it to match your new code, task now uses spawned_entities attribute from Bridges

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: okay, maybe i will define task more clearly, for example - "all cubes have to be adjacent at least to one other cube" or "place all cubes in one place/cluster"

jmatejcz: also the naming  be confusing ex. displaced objects, it is counted as ->  "when the object which was in the correct place at the start, is in a incorrect place at the end", but the naming suggest an action was taken to move it

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: done

src/rai_bench/o3de_test_bench/main.py
jmatejcz: done

jmatejcz:
found new issue with getting object position from simulation, part of it was due to changes introduced in feat/benchmarking , resolved here: https://github.com/RobotecAI/rai/pull/415/commits/702bb00bd08a3c61364b46a35439ddab7200781c

Now the transform is good, but the position of object is always same as starting position, even when object was moved on simulation

jmatejcz:
> found new issue with getting object position from simulation, part of it was due to changes introduced in feat/benchmarking , resolved here: [702bb00](https://github.com/RobotecAI/rai/commit/702bb00bd08a3c61364b46a35439ddab7200781c)
> 
> Now the transform is good, but the position of object is always same as starting position, even when object was moved on simulation

fixed here: https://github.com/RobotecAI/rai/commit/871f707347b171b4725a5c3daac26ef4b441e7cb

rachwalk (changes requested):
When attempting to run the according to PR description I get an error:

```
Traceback (most recent call last):
  File "/home/krachwal/projects/internal/rai/src/rai_bench/rai_bench/main.py", line 21, in <module>
    from rai_bench.benchmark_model import (
ModuleNotFoundError: No module named 'rai_bench'
```

It seems that poetry install doesn't install the new package.

jmatejcz:
> When attempting to run the according to PR description I get an error:
> 
> ```
> Traceback (most recent call last):
>   File "/home/krachwal/projects/internal/rai/src/rai_bench/rai_bench/main.py", line 21, in <module>
>     from rai_bench.benchmark_model import (
> ModuleNotFoundError: No module named 'rai_bench'
> ```
> 
> It seems that poetry install doesn't install the new package.

added package to pyproject.toml

jmatejcz:
the scenario should not run until all services are up ( case when grounding dino is downloading weights)

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,246 @@
...
+        scenarios = []
+        for task in tasks:
+            for sim_conf in simulation_configs:
+                try:
+                    scenarios.append(Scenario(task=task, simulation_config=sim_conf))
+                except ValueError as e:
+                    print(
+                        f"Could not create Scenario from task: {task.get_prompt()} and simulation_config: {sim_conf}, {e}"
+                    )
+        return scenarios
```
MagdalenaKotynia: What is the purpose of this method? What does this try-except block handle?

src/rai_sim/rai_sim/o3de/o3de_bridge.py
```diff
@@ -304,3 +308,57 @@ def from_ros2_pose(self, pose: Pose) -> PoseModel:
...
+        request.target_pose.pose.position.x = pose.translation.x
+        request.target_pose.pose.position.x = pose.translation.y
+        request.target_pose.pose.position.z = pose.translation.z
+
+        if pose.rotation:
+            request.target_pose.pose.orientation.x = pose.rotation.x
+            request.target_pose.pose.orientation.y = pose.rotation.y
+            request.target_pose.pose.orientation.z = pose.rotation.z
+            request.target_pose.pose.orientation.w = pose.rotation.w
+        else:
```
MagdalenaKotynia: For what is this default rotation of 180 degrees in the x-axis?

src/rai_bench/rai_bench/main.py
```diff
@@ -0,0 +1,174 @@
...
+    tasks = [
+        GrabCarrotTask(logger=bench_logger),
+        PlaceCubesTask(logger=bench_logger),
+    ]
+    scenarios = Benchmark.create_scenarios(
+        tasks=tasks, simulation_configs=simulations_configs
+    )
+
+    # custom request to arm
+    base_arm_pose = PoseModel(translation=Translation(x=0.5, y=0.1, z=0.3))
```
MagdalenaKotynia: Please add rotation to the base arm pose. The base arm pose seems to be rotated 180 degrees in the x-axis.

src/rai_bench/rai_bench/benchmark_model.py
MagdalenaKotynia: > okay, maybe i will define task more clearly, for example - "all cubes have to be adjacent at least to one other cube" or "place all cubes in one place/cluster"

It is not a problem in task definition but in metric calculation. A more precise prompt will not resolve the problem that I described above (Maybe the agent will behave better but we cannot measure it properly with the current metric. 





> also the naming be confusing ex. displaced objects, it is counted as -> "when the object which was in the correct place at the start, is in a incorrect place at the end", but the naming suggest an action was taken to move it

Actually, the `correct place` phase is still confusing because there is no one correct place for the cubes in this task. The success of this task depends on the relation between the cubes' placement.

src/rai_bench/rai_bench/o3de_test_bench/tasks/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from rai_bench.o3de_test_bench.tasks.grab_carrot_task import GrabCarrotTask
+from rai_bench.o3de_test_bench.tasks.place_cubes_task import PlaceCubesTask
+
+__all__ = ["GrabCarrotTask", "PlaceCubesTask"]
```
MagdalenaKotynia, line 17: What is the purpose of using __ all __ here?

src/rai_bench/rai_bench/o3de_test_bench/tasks/grab_carrot_task.py
```diff
@@ -0,0 +1,99 @@
...
+    SimulationBridge,
+)
+from rai_sim.simulation_bridge import SimulationConfig
+
+
+class GrabCarrotTask(Task):
+    def get_prompt(self) -> str:
+        return "Manipulate objects, so that all carrots to the left side of the table (positive y)"
+
+    def validate_scene(self, simulation_config: SimulationConfig) -> bool:
```
MagdalenaKotynia: I don't catch the sense of this method. It only verifies if the "carrot" prefab is present in the SimulationConfig. This check does not guarantee that the prefab named "carrot" is present in the scene. This make the implementation of this method inconsistent with the docstring of abstract method. 
Aprat from that, IMO checking if the specific prefab is available, should be handled by the setup_scene method of the SimulationBridge. On the example of O3DExROS2Bridge, if the prefab named "carrot" will not be available for specific scene, the _spawn_entity method will throw [error](https://github.com/RobotecAI/rai/blob/929ca5c1832f1460f0d602031ffadcf2bf5563e6/src/rai_sim/rai_sim/o3de/o3de_bridge.py#L158).

Therefore, I suggest removing the validate_scene method.

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,246 @@
...
+            for j, p2 in enumerate(positions):
+                if i != j:
+                    if self.is_adjacent(p1, p2, threshold_distance):
+                        adjacent_count += 1
+                        break
+
+        return adjacent_count
+
+
+class Scenario:
```
MagdalenaKotynia: What was the purpose of converting the BaseModel class to the standard class? 

Apart from my suggestion to remove the validate_scene method in the previous comment, such validation could be done in the BaseModel class (see Pydantic [validators](https://docs.pydantic.dev/2.0/usage/validators/))

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,246 @@
...
+                f"TASK SCORE: {result}, TOTAL TIME: {total_time:.3f}, NUM_OF_TOOL_CALLS: {tool_calls_num}"
+            )
+
+            self.results.append(
+                {
+                    "task": scenario.task.get_prompt(),
+                    "initial_score": initial_result,
+                    "final_score": result,
+                    "total_time": f"{total_time:.3f}",
+                    "numer_of_tool_calls": tool_calls_num,
```
MagdalenaKotynia:
```suggestion
                    "number_of_tool_calls": tool_calls_num,
```

jmatejcz:
> the scenario should not run until all services are up ( case when grounding dino is downloading weights)

propsed solution in https://github.com/RobotecAI/rai/pull/426/commits/ac14185b84811e57850665aaa8938540857ac68b

.gitignore
```diff
@@ -172,3 +172,6 @@ logs/
 
 src/examples/*-demo
 artifact_database.pkl
+
+imgui.ini
+o3de_config.yaml
```
MagdalenaKotynia: This file is not included in the repo (what is good) and it is user's specific config file, so I think that it should not be added to .gitignore.

MagdalenaKotynia (changes requested):
I left some comments. Apart from that, there is quite a lot of missed/wrong typing. Could you please fix it and consider switching to a more strict typing checking mode?

src/rai_bench/rai_bench/o3de_test_bench/tasks/grab_carrot_task.py
MagdalenaKotynia: Update after private discussion: change name and docstring to reflect functionality of the method.

.gitignore
jmatejcz: removed config from gitignore

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: applied

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: scenario is validating if task can be paired with scene, if not ValueError is raised

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: as we talked in the office, the task if defined differently now so this issue is no longer present

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: the purpouse was to validate in constructor, apart from that it makes more sense to me that Scenario is normal class, not a pydatnic model

src/rai_bench/rai_bench/o3de_test_bench/tasks/grab_carrot_task.py
jmatejcz: renamed function and removed example from docstring in abstract Task class

src/rai_bench/rai_bench/o3de_test_bench/tasks/__init__.py
jmatejcz: it is here so the imports of tasks are easier - oneline, instead of separate import for every task

src/rai_sim/rai_sim/o3de/o3de_bridge.py
jmatejcz: deleted default rotation, added this rotation to main.py

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,249 @@
...
+
+from rai.messages import HumanMultimodalMessage
+from rai_sim.simulation_bridge import (
+    PoseModel,
+    SimulationBridge,
+    SimulationConfig,
+    SpawnedEntity,
+)
+
+SimulationBridgeT = TypeVar(
```
MagdalenaKotynia: I see that this is not used anywhere, please remove it.

src/rai_sim/rai_sim/o3de/o3de_bridge.py
```diff
@@ -199,6 +205,35 @@ def get_scene_state(self) -> SceneState:
             )
         return SceneState(entities=entities)
 
+    def _is_robotic_stack_ready(
+        self, simulation_config: O3DExROS2SimulationConfig
+    ) -> bool:
+        retries = 0
```
MagdalenaKotynia: Please parametrize number of retires with some default value

src/rai_sim/rai_sim/o3de/o3de_bridge.py
```diff
@@ -199,6 +205,35 @@ def get_scene_state(self) -> SceneState:
...
+                all(srv in service_names for srv in simulation_config.required_services)
+                and all(tp in topics_names for tp in simulation_config.required_topics)
+                and all(
+                    ac in service_names for ac in simulation_config.required_actions
+                )
+            ):
+                self.logger.info("All required services are available.")
+                return True
+
+            time.sleep(5)
```
MagdalenaKotynia, line 233: Is such big time sleep needed?

src/rai_bench/rai_bench/benchmark_model.py
```diff
@@ -0,0 +1,249 @@
...
+        self.results: List[Dict[str, Any]] = []
+        if logger:
+            self._logger = logger
+        else:
+            self._logger = logging.getLogger(__name__)
+
+    @classmethod
+    def create_scenarios(
+        cls, tasks: List[Task], simulation_configs: List[SimulationConfig]
+    ) -> list[Any]:
```
MagdalenaKotynia: Why is it typed to return list[Any]?

src/rai_sim/rai_sim/o3de/o3de_bridge.py
jmatejcz: the download of the model and other setups can last a long time, no point in checking more frequently IMO

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: deleted it

src/rai_bench/rai_bench/benchmark_model.py
jmatejcz: you already fixed it, thanks ;)

src/rai_sim/rai_sim/o3de/o3de_bridge.py
jmatejcz: done

rachwalk (approved):
LGTM

Issue #427, opened 2025-02-19
Issue with calling service spawn_entity after some successful attempts.


MagdalenaKotynia:
**Describe the bug**
The issue related to rai_sim package, commit: f243525d77f0ee55171168e48be0d8ea21033e76
When running scene_setup method of O3DExROS2Bridge in a loop the following error occurs after some successful attempts:
```
Exception in thread Thread-1 (spin):
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 294, in spin
    self.spin_once()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 794, in spin_once
    self._spin_once_impl(timeout_sec)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 791, in _spin_once_impl
    future.result()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/task.py", line 94, in result
    raise self.exception()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/task.py", line 239, in __call__
    self._handler.send(None)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 430, in handler
    arg = take_from_wait_list(entity)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 354, in _take_subscription
    with sub.handle:
rclpy._rclpy_pybind11.InvalidHandle: cannot use Destroyable because destruction was requested
```
The number of attempts after the error occurs is different each time.

The errors occurs [here](https://github.com/RobotecAI/rai/blob/4c880e8a354d869b282d005536d1f8c145acde77/src/rai_core/rai/communication/ros2/api.py#L438) when 
 [trying to call service spawn_entity](https://github.com/RobotecAI/rai/blob/4c880e8a354d869b282d005536d1f8c145acde77/src/rai_sim/rai_sim/o3de/o3de_bridge.py#L147) and is not being catched through try-except (when trying to do so).

**To Reproduce**

1. Go to f243525d77f0ee55171168e48be0d8ea21033e76
[...]

PR #428, opened 2025-02-20
Fix tf listener in rai_sim module

pijaro:
## Purpose

Fixes ROS 2 executor random crashes. 

## Proposed Changes

Move TF buffer to node scope.

## Issues

- https://github.com/RobotecAI/rai/issues/427

## Testing

- Following the procedure described in https://github.com/RobotecAI/rai/issues/427


MagdalenaKotynia:
@pijaro Thank you for this fix! I tested it and the issue no longer occurs.  

@maciejmajek @rachwalk Since I am not very familiar with connectors, I will leave the code review to someone more knowledgeable on the subject.

maciejmajek (approved):
I think this actually makes sense. Can't test it until Monday, but i believe it should work as intended/better with this change

rachwalk (approved):
LGTM, merging

PR #449, opened 2025-03-07
refactor: demos rewrite, rai_interfaces import guards

maciejmajek:
## Purpose 

Using RAI as a standalone framework is unfeasible due to many custom ROS 2 dependencies. Some of the parts of RAI could already be used in a pip install rai installation.

## Proposed Changes

This PR adds various changes including:
- Guarding the rai_interfaces, rai_openset ... imports
- Removing deprecated files
- Rewriting demos

## Testing


boczekbartek:
@maciejmajek Could you please resolve confilicts?

src/rai_core/rai/node.py
boczekbartek, line 1: Removing this file will break [turtlebot_demo](https://github.com/RobotecAI/rai/blob/9564f4fb5fc23e78418070a80f8ac652382e1f29/src/examples/turtlebot4/turtlebot_demo.py#L73), so I think it should be refactored as well

examples/rosbot-xl-demo.py
```diff
@@ -114,32 +81,31 @@ def main(allowlist: Optional[Path] = None):
     (0.92, 1.01, 0.0)
     """
 
-    node = RaiStateBasedLlmNode(
```
boczekbartek, line 117: @maciejmajek 
I can see that `RaiStateBasedLlmNode` has been replaced with a simpler agent.
Could you please clarify which of it's capabilities are planned to be added again and which will be dropped? Here is the overview of features:
- `robot state` 
      - observing a ros2 topic and postprocessing/interpreting it's output. Adding the output of interpretation to the agent history.
      - observing rosout with `rai_state_logs` node
- `ros2 action interface` - this node had ros2 action interface, which could be used to send a mission to the robot. Then summary of langgraph agent has been posted as ros2 action feedbacks

src/rai_whoami/rai_whoami/rai_whoami_node.py
boczekbartek, line 1: Could you please clarify which features of rai_whoami are planned to be still supported in rai?
- RAG for documentation
- "embodiment" in system prompt using documentation and images
- setting robot constitution

boczekbartek (commented):
@maciejmajek 
This PR introduces major changes. I left some initial questions about the plan support rai features.

examples/rosbot-xl-demo.py
maciejmajek: > robot state

I think we can reintroduce it either by using a separate node in a graph, or by specifying topic name in connector sources. 

> ros2 action interface

The action interface can be reintroduced with support for actions in connectors. If you feel that this is important, we can add it to examples catalog showcasing ros2 usage using action server api from rclpy

src/rai_whoami/rai_whoami/rai_whoami_node.py
maciejmajek: I don't think that the system prompt propagation should be done via ros2 or any other communication protocol at the moment. Right now I find it easier to use text files, but in the future we could build a specific database with system prompt versioning.

examples/rosbot-xl-demo.py
boczekbartek: > I think we can reintroduce it either by using a separate node in a graph, or by specifying topic name in connector sources.

I think it'll require both, because state should be received (connector) and then placed in the LLM context (node in the graph). In general, it sounds like a good idea, if you don't plan to refactor it in this PR, how about adding an Issue that has to be resolved before the next release to add it again?

src/rai_whoami/rai_whoami/rai_whoami_node.py
boczekbartek: - I think the design can be changed to the easier one (for example text files instead of ros2 node).
However my question is more about removing/leaving functionalities:
- automatic creation of the system prompt using documents and images (can be used to create a system prompt in a file)
- RAG for robot documentation (I think it might be too much for the `rai_core`, so I think `rai_whoami` package might be a good place for it. ros interface is up to discussion, probably RAG can be queried directly in the tool call.)

maciejmajek: rai whoami issue https://github.com/RobotecAI/rai/issues/470

examples/rosbot-xl-demo.py
maciejmajek: @coderabbitai please create an issue regarding action and service servers in BaseConnector API

boczekbartek: @maciejmajek how about the concept of `robot_state`?

src/rai_core/rai/node.py
boczekbartek: bumping this thread

boczekbartek:
@maciejmajek I tried:
```
rm -rd build install log
```

```python
import rai
```

and I have this output:
```
ModuleNotFoundError: No module named 'rai_interfaces'
```

I think the purpose of this PR is to have this error only in ros-dependent rai packages like demos, right?

examples/rosbot-xl-demo.py
maciejmajek: @coderabbitai please create an issue regarding parsing the environment state using a predefined mechanism similar to https://github.com/RobotecAI/rai/blob/210a066a2b35579b678183c202dba38e3031e64c/src/rai/rai/node.py#L560-L591 but in the RAI 2.0 architecture using connectors

src/rai_core/rai/node.py
boczekbartek: decided to work on it under: https://github.com/RobotecAI/rai/issues/476

boczekbartek (approved):
@maciejmajek 
- `rai_interfaces` doesn't break `rai` import, warning is shown
- tested code for `agriculture` and `rosbot` demos: code doesn't throw errors, but doesn't work. Refactoring these demos will be tracked under issues: https://github.com/RobotecAI/rai/issues/477, https://github.com/RobotecAI/rai/issues/474
- `turtlebot demo` needs refactor: https://github.com/RobotecAI/rai/issues/476
- `manipulation demo` works 👍🏼 

approving, and let's move forward with demos under issues

PR #450, opened 2025-03-07
feat: react agent

maciejmajek:
## Purpose

Rewrite: simplest agent built on top of the BaseAgent and HRIConnector api. 

## Proposed Changes

SimpleAgent with input and output gathered/sent via new connector api. 

## Issues

- Links to relevant issues

## Testing
Terminal 1
```bash
python examples/agents/react.py
```
Terminal 2
```bash
ros2 topic echo /to_human rai_interfaces/msg/HRIMessage
```
Terminal 3
```bash
ros2 topic pub /from_human rai_interfaces/msg/HRIMessage "{'text': 'Introduce yourself if 100 words'}" --once
```
> [!NOTE]  
> The message may not be received on the first try due to asynchronous nature of ROS 2 discovery mechanism.

## Summary by CodeRabbit

- **New Features**
  - Introduced a standalone ROS2 agent script to enable simple agent operations.
  - Launched new frameworks for language model–driven agents, supporting tool integration and real-time token processing.
  - Enhanced message handling by queuing topic-based messages for smooth asynchronous processing.

- **Refactor**
  - Improved default configurations for logging and communication parameters for increased stability.
  - Streamlined message evaluation to ensure reliable conditional processing.


maciejmajek:
@coderabbitai full review

src/rai_core/rai/agents/langchain/callback.py
```diff
@@ -0,0 +1,76 @@
...
+    ):
+        self.connectors = connectors
+        self.aggregate_chunks = aggregate_chunks
+        self.splitting_chars = splitting_chars or ["\n", ".", "!", "?"]
+        self.chunks_buffer = ""
+        self.max_buffer_size = max_buffer_size
+        self._buffer_lock = threading.Lock()
+
+    def _should_split(self, token: str) -> bool:
+        return any(token in char for char in self.splitting_chars)
```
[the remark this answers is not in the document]
maciejmajek: Refactored

src/rai_core/rai/agents/langchain/callback.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def _send_all_targets(self, token: str):
+        for connector_name, connector in self.connectors.items():
+            connector.send_all_targets(AIMessage(content=token))
+            logger.debug(f"Sent token to {connector_name}")
+
```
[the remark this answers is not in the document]
maciejmajek: I don't see that happening, but it won't hurt

boczekbartek (changes requested):
@maciejmajek 
I've tested the `ReActAgent` using provided commands and I had some commumication issues - the agent sometimes skips the messages on `/from_human` topic.
I'm on `ubuntu22.04`, `ros2 humble`, `rmw_fastrtps_cpp`

See video below:
https://github.com/user-attachments/assets/85083bb9-83e6-4205-9087-0f4a2126329a

boczekbartek (changes requested):
I tested again after last fix and it seems like the previously reported bug is fixed, thanks! 
But I noticed a other unexpected behaviour of the agent.
The output looks suspicious, because `/from_human` topic is echoed to the `/to_human` and I am often not getting the correct response from LLM.

Here was my request:
```
for i in $(seq 1 20); do echo $i; ros2 topic pub /from_human rai_interfaces/msg/HRIMessage "{'text': \"Say rai $i times\"}" --once; done
```
And here are output logs from the agent and log from `/to_human`
[react.3.log](https://github.com/user-attachments/files/19237724/react.3.log)
[to_human.3.log](https://github.com/user-attachments/files/19237725/to_human.3.log)

The agent received correctly 20 messages (which proves the the previous bug was fixed)
```
$ grep Received react.3.log | wc -l
      20
```

But can't see the answer to any of the requests to "Say rai 11-19 times".

I was using `gpt-4o-mini` for the logs, but also tried `gpt-4o` with similar result.

@maciejmajek what do you think?

maciejmajek:
@coderabbitai please add an issue regarding ReActAgent's behavior on new message during agent's invocation. At the moment, the messages are queued but there are other possibilities that are equally important: 
- dropping the current invoke, to respond to the new data as quickly as possible 
- dropping the new message

boczekbartek:
@maciejmajek 
I've tested again with [the same example](https://github.com/RobotecAI/rai/pull/450#pullrequestreview-2683505481) with repeating rai and it worked well this time. Thank you for fixing! 

I've also tested with image input using a simple publisher in python (code below) and I received correct images` descriptions on `/to_human` topic!

One thing left from my side before approving is to check the implementation details of this PR.

<details><summary>code</summary>
```python
import os
import rclpy
from rclpy.node import Node
from sensor_msgs.msg import Image
import cv2
import time
from rai_interfaces.msg import HRIMessage

class ImagePublisher(Node):
    def __init__(self):
        super().__init__('image_publisher')
        self.get_logger().info('Image publisher started')

        # Create an image publisher node with the specified QoS profile
        self.publisher_ = self.create_publisher(HRIMessage, '/from_human', 10)

        # Set up the file path and name for each image to be published
        self.file_path = '/home/bboczek/Pictures'
        
    def publish_image(self):
        # Iterate over all images in the directory
        for filename in os.listdir(self.file_path):
            # Check if the file is an image (assuming images are in PNG format)
            if filename.endswith('.jpeg'):
                # Load the image from disk
                img = cv2.imread(os.path.join(self.file_path, filename))

                # Convert the image to a ROS image message
                msg = Image()
                msg.header.stamp = self.get_clock().now().to_msg()
                msg.header.frame_id = 'camera'
                msg.encoding = 'bgr8'
                msg.height = img.shape[0]
                msg.width = img.shape[1]
                msg.step = img.shape[1] * 3
                msg.data = img.tobytes()

                hri_msg = HRIMessage()
                hri_msg.text = "Please describe an image"
                hri_msg.images = [msg]

                # Publish the image
                self.get_logger().info(f'Publishing image: {filename}')

                self.publisher_.publish(hri_msg)
                time.sleep(10)


def main(args=None):
    rclpy.init(args=args)
    image_publisher = ImagePublisher()
    image_publisher.publish_image()
    try:
        rclpy.spin(image_publisher)
    except KeyboardInterrupt:
        print('Keyboard Interrupt (SIGINT)')
    finally:
        image_publisher.destroy_node()
        rclpy.shutdown()

if __name__ == '__main__':
    main()
```
</details>

boczekbartek (commented):
@maciejmajek 
I've also tested attached `ReActAgent` tools and they worked fine (agent discovered topic and described image). 
One improvement to consider is to expose `timeout_sec` to LLM in tool call. I initially tried with publisher that was sending images every 5s and agent didn't work well. I propose this improvement: https://github.com/RobotecAI/rai/pull/458

Commands / code to test (besides commands from PR description):
```bash
ros2 topic pub /from_human rai_interfaces/msg/HRIMessage "{'text': 'Find ros2 topic with images and describe 1 image from it. Images come every 5 -10 seconds, so please be patient'}" --once
```

<details> <summary> Images publisher </summary>

```python
import os
import rclpy
from rclpy.node import Node
from sensor_msgs.msg import Image
import cv2
import time
from rai_interfaces.msg import HRIMessage


class ImagePublisher(Node):
    def __init__(self):
        super().__init__('image_publisher')
        self.get_logger().info('Image publisher started')

        # Create an image publisher node with the specified QoS profile
        self.publisher_ = self.create_publisher(Image, '/images', 10)

        # Set up the file path and name for each image to be published
        self.file_path = '/home/bboczek/Pictures'
        
    def publish_image(self):
        # Iterate over all images in the directory
        for filename in os.listdir(self.file_path):
            # Check if the file is an image (assuming images are in PNG format)
            if filename.endswith('.jpeg'):
                # Load the image from disk
                img = cv2.imread(os.path.join(self.file_path, filename))

                # Convert the image to a ROS image message
                msg = Image()
                msg.header.stamp = self.get_clock().now().to_msg()
                msg.header.frame_id = 'camera'
                msg.encoding = 'bgr8'
                msg.height = img.shape[0]
                msg.width = img.shape[1]
                msg.step = img.shape[1] * 3
                msg.data = img.tobytes()

                # Publish the image
                self.get_logger().info(f'Publishing image: {filename}')

                self.publisher_.publish(msg)
                time.sleep(10)


def main(args=None):
    rclpy.init(args=args)
    image_publisher = ImagePublisher()
    while True:
        image_publisher.publish_image()

if __name__ == '__main__':
    main()
```
</details>

src/rai_core/rai/agents/react_agent.py
```diff
@@ -0,0 +1,103 @@
...
+                self.logger.info("Received messages")
+                reduced_message = self._reduce_messages(received_messages)
+                langchain_message = reduced_message.to_langchain()
+                self.state["messages"].append(langchain_message)
+                # callback is used to send messages to the connectors
+                self.agent.invoke(
+                    self.state,
+                    config={"callbacks": [self.callback, *self.tracing_callbacks]},
+                )
+            time.sleep(0.3)
```
boczekbartek, line 72: Why is this sleep needed?

src/rai_core/rai/communication/ros2/api.py
```diff
@@ -391,6 +394,11 @@ class ConfigurableROS2TopicAPI(ROS2TopicAPI):
     def __init__(self, node: rclpy.node.Node):
         super().__init__(node)
         self._subscribtions: dict[str, rclpy.node.Subscription] = {}
+        self.callback_group = ReentrantCallbackGroup()
+        self.topic_msg_queue: Dict[str, Queue[Any]] = defaultdict(Queue)
+
+    def _generic_callback(self, topic: str, msg: Any):
+        self.topic_msg_queue[topic].put(msg)
```
boczekbartek: I think max queue size should be set to `self.topic_msg_queue`. If we attach to the publisher that sends loads of data,  this can quickly cause memory problems.

boczekbartek (changes requested):
@maciejmajek I left some comments to the code

src/rai_core/rai/agents/react_agent.py
maciejmajek: Reducing cpu usage, allowing a little bit more time for other sources to publish data

src/rai_core/rai/communication/ros2/api.py
maciejmajek: What should be the default behavior? Drop oldest or newest?

boczekbartek: Hmm, in ros2 it depends on QoS settings. How it should be implemented `Ros2Connector` unfortunately I'm not sure...
I'll think about it

src/rai_core/rai/agents/langchain/react_agent.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    ai_msg = llm.invoke(state["messages"])
+    state["messages"].append(ai_msg)
+
+
+def create_react_agent(
+    llm: Optional[BaseChatModel] = None, tools: Optional[List[BaseTool]] = None
+) -> Runnable[ReActAgentState, ReActAgentState]:
+    """Create a react agent that can process messages and optionally use tools.
```
rachwalk, line 69: I am not entirely certain as to the naming convention of this function (and file, that's due to conflict with the agent named in the same way).
My issues are:
- a ReAct agent makes no sense if tools are None (can't act)
- this function creates a langchain runnable, not an agent (in RAI nomenclature)

In fact from what I see this function is just syntactic sugar for creating a most basic optionally tool-bound langchain runnable. It definitely doesn't allow for creating more complex langgraph structures, which could be used to create a more sophisticated ReAct agent.

In general design terms in OOP this should correspond to a LangchainRunnableFactory, with a method for creation of a basic runnable like that -- but this would go beyond the scope of this PR.

Within a scope of this PR however, this file and all of its functions should be renamed to accurately reflect what functionality is provided (not how this functionality is used within the scope of this PR). Accordingly to the `callback.py` I would consider renaming this file `runnable.py`, and rename functions accordingly. Then I'd consider the issue of how to abstract langchain runnable functionality into factory further down the line.

src/rai_core/rai/agents/react_agent.py
```diff
@@ -0,0 +1,103 @@
...
+        self, received_messages: Dict[str, Dict[str, HRIMessage]]
+    ) -> HRIMessage:
+        hri_payload = HRIPayload(text="", images=[], audios=[])
+        for connector_name, connector_sources in received_messages.items():
+            hri_payload.text += f"{connector_name}\n"
+            for source_name, source_message in connector_sources.items():
+                hri_payload.text += f"{source_name}: {source_message.text}\n"
+                hri_payload.images.extend(source_message.images)
+                hri_payload.audios.extend(source_message.audios)
+        return HRIMessage(payload=hri_payload, message_author="human")
```
rachwalk, line 103: This functionality seems like it could be highly reusable, I'd consider moving it to base Agent.

src/rai_core/rai/communication/ros2/api.py
rachwalk: If `Queue`'s `max_size` is set `put` will block until space becomes available. This would likely cause callback stacking and also cause memory issues. `put_nowait` will raise `Queue.Empty`. Given the that, I would opt for there being no default `max_size`, however if `max_size` is specified (optionally), the user should be required to specify overflow policy - `"drop_first"` or `"drop_last"`. That gives an easy and explicit fix for users who run into memory issues, but doesn't cause unexpected dropped messages for users who never encounter that.

src/rai_core/rai/agents/langchain/react_agent.py
```diff
@@ -0,0 +1,109 @@
...
+        If state is invalid or LLM processing fails
+    """
+
+    ai_msg = llm.invoke(state["messages"])
+    state["messages"].append(ai_msg)
+
+
+def create_react_agent(
+    llm: Optional[BaseChatModel] = None, tools: Optional[List[BaseTool]] = None
+) -> Runnable[ReActAgentState, ReActAgentState]:
```
rachwalk: I am not entirely certain as to the naming convention of this function (and file, that's due to conflict with the agent named in the same way).
My issues are:
- a ReAct agent makes no sense if tools are None (can't act)
- this function creates a langchain runnable, not an agent (in RAI nomenclature)

In fact from what I see this function is just syntactic sugar for creating a most basic optionally tool-bound langchain runnable. It definitely doesn't allow for creating more complex langgraph structures, which could be used to create a more sophisticated ReAct agent.

In general design terms in OOP this should correspond to a LangchainRunnableFactory, with a method for creation of a basic runnable like that -- but this would go beyond the scope of this PR.

Within a scope of this PR however, this file and all of its functions should be renamed to accurately reflect what functionality is provided (not how this functionality is used within the scope of this PR). Accordingly to the `callback.py` I would consider renaming this file `runnable.py`, and rename functions accordingly. Then I'd consider the issue of how to abstract langchain runnable functionality into factory further down the line.

src/rai_core/rai/agents/react_agent.py
rachwalk, line 103: This seems like general functionality, consider moving it to BaseAgent.

rachwalk (changes requested):
I was able to run and test a solution, good job. I have however some major issues with the naming conventions introduced with this PR. Please consider renaming the relevant files.

src/rai_core/rai/agents/langchain/react_agent.py
maciejmajek: >  which could be used to create a more sophisticated ReAct agent.

ReAct is a term used in the Lang* ecosystem to describe one of the simplest agent structures.

> A ReAct agent makes no sense if tools are None (can't act).

That's true. The tools are optional, making it reusable in the s2s use case. Writing DRY code within LangGraph is actually quite challenging. I believe our s2s or any other LLM agent will be used with tools, making this exception justifiable despite the naming inconsistency.

Regarding the naming, I agree.
I think

```python
create_react_runnable
```
should suffice.

src/rai_core/rai/communication/ros2/api.py
maciejmajek: [b3ac9b3](https://github.com/RobotecAI/rai/pull/450/commits/b3ac9b3ac93220fdd77a6a3c06f0c87edd25e283)

src/rai_core/rai/agents/langchain/react_agent.py
maciejmajek: [9e23794](https://github.com/RobotecAI/rai/pull/450/commits/9e237943465ee908ccbbb0f6476dbfee47444715)

src/rai_core/rai/agents/react_agent.py
maciejmajek: Not sure if polluting the BaseAgent with the HRI namespace would provide any benefits.

src/rai_core/rai/communication/ros2/api.py
boczekbartek: > the user should be required to specify overflow policy - "drop_first" or "drop_last"

Good idea! Thanks @maciejmajek for implementing!

> That gives an easy and explicit fix for users who run into memory issues

I would set `drop_first` policy as a default, not to have the memory issue as a default behaviour.

rachwalk: @boczekbartek the memory issue only arises in case of high-publishing-frequency, low-consumption rate situations, where `max_size` is not specified. Setting `drop_first` as the default policy, would also require setting a default `max_size`, which would generally be highly dependant on particular user setup (as well as relative frequencies of publishing and consumption of messages), and while a generally high value (i.e. `65536`) could be acceptable, having the default behaviour being dropping messages is something I would personally find confusing.

That being said, if @maciejmajek thinks it's best to have a default `max_size` (with an option to remove it), then I would heavily reccomend that `logger.warning` is added where the message is dropped. If `drop_first` is the default, than the user should be explicitely informed that it's RAI logic dropping the message (and not for example bad QoS on ROS2 setup).

boczekbartek: @rachwalk 
In current case [ConfigurableROS2TopicsAPI](https://github.com/RobotecAI/rai/blob/08a75029ddc9d50a3362698b2c1f541670c129a5/src/rai_core/rai/communication/ros2/api.py#L398) is only used in [ROS2HRIConnector](https://github.com/RobotecAI/rai/blob/08a75029ddc9d50a3362698b2c1f541670c129a5/src/rai_core/rai/communication/ros2/connectors.py#L195). In that case I agree that high-publishing-frequency, low-consumption rate situations is probably quite rare, because messages comes from human. 
The question is whether `ConfigurableROS2TopicsAPI` will be used as a tool call or in a [robot state](https://github.com/RobotecAI/rai/pull/449#discussion_r2000951279). If so I can imagine the situation where it's used to subscribe to high frequency topics with loads of data (pointcloud/cameras) and queue footprint will grow significantly, because LLM will check infrequently for messages or state is read with low frequency. I think it's a very probable usage of `ConfigurableROS2TopicsAPI` in `rai` so I would opt for `safe` default behavior and being able to configure it to be unsafe. Adding warning is fine.
I think now it's up to @maciejmajek decision about which intuition is more convincing.

boczekbartek (approved):

rachwalk (approved):

src/rai_core/rai/communication/ros2/api.py
maciejmajek: @coderabbitai Please add an issue regarding the problem discussed in this thread.

PR #458, opened 2025-03-14
feat(`ros2_tools`): add `timeout_sec` param to `get_ros2_image`

boczekbartek:
## Purpose

Current implementation makes it possible only to get images from ros2 topics with frequency >= 1Hz. Which might not always be the case.

## Proposed Changes

Make `timeout_sec` parameter available to the LLM in tool call.


maciejmajek:
Please provide a default value. 1s should be enough

boczekbartek:
@maciejmajek feedback applied, tests pass, rebased. I think it's ready to merge.

maciejmajek (approved):

PR #460, opened 2025-03-17
feat: add ground truth object positions tool

knicked:
## Purpose

Provide a tool that returns ground truth object positions to simplify agent benchmarking without openset vision tools.

## Proposed Changes

- Added the tool that returns the positions of objects given a correct prefab name using a simple word matching criterion.
- Replace the tool with the new one in `o3de_test_benchmark.py`

## Testing
Tested by running `python src/rai_bench/rai_bench/examples/o3de_test_benchmark.py`


src/rai_core/rai/tools/ros/manipulation.py
```diff
@@ -170,7 +176,24 @@ def _run(self, object_name: str):
...
 
-        if len(mani_frame_poses) == 0:
-            return f"No {object_name}s detected."
-        else:
-            return f"Centroids of detected {object_name}s in manipulator frame: [{', '.join(map(self.format_pose, mani_frame_poses))}]. Sizes of the detected objects are unknown."
+        return self.format_message(mani_frame_poses)
+
+
+class GetObjectPositionsGroundTruthTool(GetObjectPositionsTool):
+    simulation: SimulationBridge
```
jmatejcz, line 183: please adjust type hinting, i recommend using at least standard type checking mode in pylance which is integrated into vscode

src/rai_core/rai/tools/ros/manipulation.py
```diff
@@ -170,7 +176,24 @@ def _run(self, object_name: str):
...
+
+    @staticmethod
+    def format_pose(pose: Pose):
+        return f"Centroid(x={pose.translation.x:.2f}, y={pose.translation.y:2f}, z={pose.translation.z:2f})"
```
jmatejcz: Does Pose has translation attribute? please adjust type checking here. You can also add return type hints to this class

jmatejcz:
looks and works fine, I only suggested adjusting type hinting

maciejmajek (commented):
Well done.
One comment:
I think that this tool should be included in the rai_sim package, as it's simulation-specific.

src/rai_core/rai/tools/ros/manipulation.py
knicked: I added a correct type here after moving the tool to the `rai_sim` package.

knicked:
I have moved the tool to the `rai_sim` package, inside a new file `tools.py`

src/rai_sim/rai_sim/tools.py
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from rai.tools.ros.manipulation import GetObjectPositionsTool
+
+from rai_sim.simulation_bridge import Pose, SimulationBridge
+
+
+class GetObjectPositionsGroundTruthTool(GetObjectPositionsTool):
```
maciejmajek: I don't think it's necessary to inherit one tool from another, as it ends up making you initialize the GT tool with stuff that's pretty much useless.

Please refactor this into a separate tool instead.

maciejmajek (changes requested):

src/rai_sim/rai_sim/tools.py
knicked: The point of inheritance here was to have the ground truth tool inherit the name, description and output format of the original tool, so it's indistinguishable for the agent from the original tool. I think the unused parameters could be initialized with `None`s so we don't have to create them if it's expensive.

knicked: I came up with a way to achieve this without inheritance, hence getting rid of those unnecessary parameters. The tool has been refactored.

maciejmajek: Importing the tool from rai to rai_sim imposes a series of limitations:
- openset must be installed in order (which is an optional package)
- openset's dependencies must be installed

Additionally, the benchmark relies on the fact, that parts of the framework are constant. In other words, implementing the tool from scratch inside the rai_sim package could yield an additional value to the benchmark. 
We should try to untangle the dependencies at the cost of redundancy: just copy the tool, remove unused parts and voila :). If you think otherwise, please comment.

knicked: I understand your point; my intuition was that copying the code is not a good practice, but since you say that it's beneficial to have that tool be independent of the `rai` package, then I agree with you and I refactored the tool again, this time implementing it from scratch.

I squashed the previous commits to maintain clean history of the files in `rai` package.

maciejmajek (approved):
LGTM

Issue #63, opened 2024-07-04
SIM: custom ros2 interfaces benchmark


maciejmajek:
Extend the tool calling benchmark to custom ros2 interfaces, e.g. rai_interfaces. 

---
Implementation:
- 1 topic that RAI can subscribe to (get msg tool)
- 1 topic with subscriber (publish msg tool)
- 1 service (call service tool)
- 1 action (start action tool)
The implementation of ros2 node should be mocked. The benchmark should validate the arguments passed to the tools (if it's compatible with the custom message definition)

jmatejcz:
try 1st test:

- [x] mock PublishROS2MessageTool
- [x] mock GetROS2MessageInterfaceTool
- [x] task to publish message in topic, should check if agent checked message interface and then properly published it

jmatejcz:
- [x] adjust string output of  GetROS2MessageInterfaceTool to be formatted like real output
- [x] return exact fail message when something goes wrong in mock tools
- [x] move verification of specific task like message published to the verify_tool_calls_method

jmatejcz:
the issue occured: https://github.com/RobotecAI/rai/issues/492

jmatejcz:
mocks and tests for:
services:
- [x] CallROS2ServiceTool mock
- [x] task to call custom service

actions:
- [x] StartROS2ActionTool mock 
- [x] GetROS2ActionFeedbackTool mock 
- [x] GetROS2ActionResultTool mock 
- [x] CancelROS2ActionTool mock 
- [x] GetROS2ActionIDsTool mock 
- [x] task to call custom action

topics:
- [x] task get custom topic

jmatejcz:
- [x] refactor `expected` variables like expected_topics into `available`

jmatejcz:
waiting for https://github.com/RobotecAI/rai/pull/491
to finish service and action tasks

jmatejcz:
tasks for other types:
- [x] AudioMessage
- [x] RAIDetectionArray
- [x] RAIGRoundedSam
- [x] RAIGroundedDino
- [x] StringList
- [x] VectorStoreRetrieval
- [x] WhatISee

jmatejcz:
refactored  interface tool in https://github.com/RobotecAI/rai/pull/487/commits/447399b56ba70c6380e508da83d8fd2b1aa34ecf

jmatejcz:
- [x] some parent class for common code for custom interfaces tasks

jmatejcz:
[...]

PR #487, opened 2025-03-27
feat: custom ros2 interfaces benchmark

jmatejcz:
## Purpose

Extend existing tool benchmark to test custom interfaces

## Proposed Changes
For now only mocks of PublishROS2MessageTool and MockGetROS2MessageInterfaceTool.
For now 1 test PublishROS2CustomMessageTask

## Issues
cause of this PR -> https://github.com/RobotecAI/rai/issues/63
issue with openai models -> https://github.com/RobotecAI/rai/issues/492
issue with get interface tool not returning types -> https://github.com/RobotecAI/rai/issues/497

## Testing
```bash
python src/rai_bench/rai_bench/examples/tool_calling_agent_test_bench.py
```
Check results in `src/rai_bench/rai_bench/experiments/tool_calling_agent_test_bench`


src/rai_bench/rai_bench/tool_calling_agent_bench/mocked_tools.py
```diff
@@ -181,3 +183,57 @@ def _run(self, object_name: str) -> str:
...
+
+        """
+        if (
+            self.expected_topic == topic
+            and self.expected_message == message
+            and self.expected_message_type == message_type
+        ):
+            return "Message published successfully"
+        else:
+            return "Failed to publish message"
```
MagdalenaKotynia: I suggest returning the same message as if the original tool returned if something goes wrong. In the future we will possibly be allowing the agent to get better after it makes a mistake in calling the tool (e.g. If the agent calls tool with wrong argument and, based on the tool output, the agent will call the tool once again with the corrected argument). Therefore, we need to reflect the behaviour of the original tool. 
Please take into account that the message can be different depending on what went wrong (e.g. when the wrong topic is passed, the message from the original tool may be different than if the wrong message_type is passed.)

src/rai_bench/rai_bench/tool_calling_agent_bench/mocked_tools.py
```diff
@@ -181,3 +183,57 @@ def _run(self, object_name: str) -> str:
...
+
+        Returns
+        -------
+        str
+            The mocked output of 'ros2 interface show' for the specified message type.
+        """
+        if msg_type in self.mock_interfaces:
+            return self.mock_interfaces[msg_type]
+        else:
+            return f"Interface for {msg_type} not found."
```
MagdalenaKotynia: The same as in https://github.com/RobotecAI/rai/pull/487#discussion_r2016470553

src/rai_bench/rai_bench/tool_calling_agent_bench/ros2_agent_tasks.py
```diff
@@ -1512,3 +1514,157 @@ def _matches_sequence(
...
+            self._check_tool_call(
+                tool_call=ai_messages[2].tool_calls[0],
+                expected_name="publish_ros2_message",
+                expected_args={
+                    "topic": "/to_human",
+                    "message": {
+                        "header": {"stamp": {"sec": 0, "nanosec": 0}, "frame_id": ""},
+                        "text": "Hello!",
+                        "images": [],
+                        "audios": [],
```
MagdalenaKotynia: I suggest making it a class attribute, as it is repeated twice in this task, so it will be easier to maintain the code.

jmatejcz:
new issue encountered, before solving that these task doesn't have much sense, as agent can build proper messages
https://github.com/RobotecAI/rai/issues/497

jmatejcz:
closing, as these changes refactored and applied here:
https://github.com/RobotecAI/rai/issues/515

PR #493, opened 2025-03-28
feat: spatial reasoning tasks

MagdalenaKotynia:
## Purpose

Tasks to test the spatial reasoning capabilities of VLMs in tool-calling agents.

## Proposed Changes

Added spatial reasoning tasks - true/false questions to the given images. 

## Issues

Need for testing the spatial reasoning capabilities of VLMs.

## Testing

python `src/rai_bench/rai_bench/examples/spatial_tool_calling_agent_bench.py`

> [!NOTE]
> The results of tasks depend not only on the spatial capabilities of the models but also on the tool-calling capabilities of the model. I suggest refactoring it in the future to make the spatial reasoning benchmark independent from the tool calling capabilities of the models.


jmatejcz:
closing as these changes refactored and applied here:
https://github.com/RobotecAI/rai/issues/515

PR #525, opened 2025-04-15
fix(`ConfigurableTopicsApi`): reading a single messaged from non configured topics

boczekbartek:
## Purpose

`ConfigurableTopicsAPI` when `receive` is called can't read 1 message from a ros2 topic. It's because code from `ROS2TopicAPI` and `ConfigurableTopicsAPI` is not fully compatible and subscribers seem not to be created properly.
By default in `ROS2TopicsAPI` the parameter `destroy_subscribers = False`. Therefore `_wait_for_message_persistent` method is selected and raises KeyError.

## Proposed Changes

- [x] add a test for the bug
- [x] parametrize ROS2TopicsAPI test with `destroy_subscribers`
- [x] fix lack of return in `ConfigurableTopicsAPI`
- [ ] fix `ConfigurableTopicsAPI`

## Testing
CI

### DIFF
--- a/src/rai_core/rai/communication/ros2/api/topic.py
+++ b/src/rai_core/rai/communication/ros2/api/topic.py
@@ -513,7 +513,8 @@ class ConfigurableROS2TopicAPI(ROS2TopicAPI):
             ValueError: For unconfigured topics, inherits parent class exceptions
         """
         if topic not in self.topic_msg_queue:
-            super().receive(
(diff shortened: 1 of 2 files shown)


boczekbartek:
@maciejmajek 
Closing this PR, because API is being refactored in https://github.com/RobotecAI/rai/pull/532

Issue #543, opened 2025-04-25
Rai bench: extracting messages from agent.stream fails


jmatejcz:
**Describe the bug**
Benchmarks collects messages from agent.stream().
Current implemention in maniupulation_o3de collects only last message every iteration.
and current implementation in tool_calling_agent collects all messages every iteration.
Both these approaches are invalid as messages can come in batches and are appended to already stored messages. This means that maniupulation_o3de can skip some messages and tool_calling_agent unnecessarily appends same messages more than once.

In manipulation O3DE it doesn't affect results as they are calculated based on end simulation positions, but can affect messages logged. In  Tool Calling Agent it can affect results.

**To Reproduce**
1. To see this bug you need to run `python src/rai_bench/rai_bench/examples/test_models.py ` in debug mode
2. Track values of event and messages in agent loop:
![Image](https://github.com/user-attachments/assets/e59ef860-401e-4bee-a9cd-3c1f314cd0d7)

**Expected behavior**
Extracted messages should be ALL UNIQUE messages returned by agent.
Tool calls extracted from these messages should also reflect all unique agent tool calls.

**Screenshots**

**Platform**

- OS: Ubuntu 22.04
- ROS 2 HUMBLE

**Version**
[Release number or commit hash.](https://github.com/RobotecAI/rai/commit/0eec253e2ead7b6b8f44e336f2a07ff28f9ad7d8)

**Additional context**

jmatejcz:
resolved here:
https://github.com/RobotecAI/rai/pull/544

PR #544, opened 2025-04-25
fix: collecting messages from agent

jmatejcz:
## Purpose
fix collecting messages from agent, described in issue linked below.
## Proposed Changes
in agent.stream() loop there is now counter for how many messages were already received and on each iteration only the [received_count: ] messages are appended.

Also fixed import issue in openset https://github.com/RobotecAI/rai/pull/544/commits/161b696d5f476e39035e861b89f71cbdb8f4f06b
which was introduced in some other refactors

## Issues 
https://github.com/RobotecAI/rai/issues/543

## Testing
1. Run  `src/rai_bench/rai_bench/examples/test_models.py` in debug mode.
2.  Track messages appended in loop 
image
Check if all unique(ids of messages) messages from state variable are appended to messages properly.

Do this for both tool calling agent and manipulaiton_o3de (change in `src/rai_bench/rai_bench/examples/test_models.py` selected benchamrk if needed).

### DIFF
--- a/src/rai_bench/rai_bench/manipulation_o3de/benchmark.py
+++ b/src/rai_bench/rai_bench/manipulation_o3de/benchmark.py
@@ -184,33 +184,36 @@ class Benchmark:
             tool_calls_num = 0
 
             ts = time.perf_counter()
+            prev_count: int = 0
             for state in agent.stream(
                 {"messages": [HumanMessage(content=scenario.task.get_prompt())]},
                 {
                     "recursion_limit": 100
                 },  # NOTE (jmatejcz) what should be recursion limit?
             ):
-                graph_node_name = list(state.keys())[0]
-                msg = state[graph_node_name]["messages"][-1]
+                node = next(iter(state))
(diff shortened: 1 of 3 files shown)


src/rai_bench/rai_bench/manipulation_o3de/benchmark.py
```diff
@@ -184,33 +184,36 @@ def run_next(self, agent: CompiledStateGraph) -> None:
             tool_calls_num = 0
 
             ts = time.perf_counter()
+            prev_count: int = 0
             for state in agent.stream(
```
maciejmajek, line 188: Why exactly are we using stream instead of invoke which returns all the messages under 'messages' key?

jmatejcz: so that we get partial results when Recursion Limit throws an error. I don't think it can  be done, when using invoke() as it return output when whole execution is over

maciejmajek (approved):

Issue #556, opened 2025-05-06
processes stay alive after manipulation


jmatejcz:
**Describe the bug**
Processes like `python run_vision_agents.py` or other from robotic stack are alive after end of execution, consuming resources. Also ollama keeps serving models for couple minutes after they are not used which results in not enough VRAM to serve next model in benchmark.

**To Reproduce**
Set up in `rai_bench/examples/test_models.py`

```python
 models_name = ["qwen2.5:7b", "llama3.2"]
    vendors = ["ollama", "ollama"]
    benchmarks = ["manipulation_o3de"]
    extra_tool_calls = [5]
    repeats = 1
```
After iteration of qwen, processes keep dangling and results in not enough resources for next iteration and crash of the program

**Expected behavior**
All processes need to end before stating net iteration in manipulation benchmark

**Platform**

- OS: Ubuntu 24.04
- ROS 2 Humble

**Version**

jmatejcz:
partially adressed here: https://github.com/RobotecAI/rai/pull/570
followup issue -> https://github.com/RobotecAI/rai/issues/571

Issue #555, opened 2025-05-05
configs in manipulation o3de benchmark not valid in new code


jmatejcz:
**Describe the bug**
After changes in https://github.com/RobotecAI/rai/pull/545, scene configs need adjustments, as they do not have `frame_id` field

**To Reproduce**
Run manipulation o3de benchmark

**Expected behavior**
proper configs loading

**Screenshots**

![Image](https://github.com/user-attachments/assets/ee22b57d-63c7-4259-804a-27ce44539dfd)

**Platform**

- OS: Ubuntu 22
- ROS 2 Humble

**Version**
Release number or commit hash.

jmatejcz:
fixed here: https://github.com/RobotecAI/rai/pull/570

PR #570, opened 2025-05-09
fix: manipulaiton o3de 

jmatejcz:
## Purpose
Manipulation o3de benchmark is not working properly after some recent PRs.

## Proposed Changes
- Changed shutdown of o3de binary and robotic stack, now after SIGINT signal does not close processes, send SIGTERM, if this does not work, KILL. This ensures closing all dandling processes after finishing benchmark.

- Redownloading weights was bugged and resulted in always redownloding weights if anything went wrong (for example cuda out of memory). On top of that this redownload created nested dir in place of previous weights which resulted in further crashes.
refactored logic to always remove file or dir before redownloding.

- Keep alive param now set to 10 seconds when running model with ollama. This prevents ollama from consuming resources after benchmark.

- Assigning "odom" as default value of frame_id for entities loaded from configs. Sending proper prefab name to spawn entity service

- Adjusted logging

- Added utils for common benchmark logic

- Catching recurssion limit error in manipulation bench

## Issues
https://github.com/RobotecAI/rai/issues/555
https://github.com/RobotecAI/rai/issues/556

## Testing
1. Run `python src/rai_bench/rai_bench/examples/test_models.py`
with settings:
```python
 models_name = ["qwen2.5:7b", "llama3.2"]
    vendors = ["ollama", "ollama"]
    benchmarks = ["manipulation_o3de"]
    extra_tool_calls = [5]
    repeats = 1
```
2. Check if all processes consuming VRAM (nvidia-smi check) are closed after qwen iteration ends.
3. Check if entites spawn correctly ( on the table )
4. Check if logs are properly stored
5. Remove weights and check if they are downloaded properly

### DIFF
--- a/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
+++ b/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
@@ -75,7 +75,8 @@ class BaseVisionAgent(BaseAgent):
             raise Exception("Could not download weights")
 
     def _remove_weights(self, path: str):
-        # NOTE (jm) sometimes downloding weights bugs and creates a dir
+        # Sometimes redownloding weights bugged and created a dir
+        # so check also for dir and remove it in both cases
         if os.path.isdir(path):
             shutil.rmtree(path)
         elif os.path.isfile(path):
--- a/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
+++ b/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
@@ -14,6 +14,7 @@
 
 
 import os
+import shutil
 import subprocess
 from pathlib import Path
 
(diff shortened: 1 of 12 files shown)


src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
```diff
@@ -72,8 +74,11 @@ def _download_weights(self, path: Path):
             self.logger.error("Could not download weights")
             raise Exception("Could not download weights")
 
-    def _remove_weights(self, path: Path):
-        if path.exists():
+    def _remove_weights(self, path: str):
+        # NOTE (jm) sometimes downloding weights bugs and creates a dir
+        if os.path.isdir(path):
+            shutil.rmtree(path)
+        elif os.path.isfile(path):
```
maciejmajek: That's interesting, I've come across the same bug today. Does it work now? Also, please use full github handle for the note

jmatejcz: The bug was appearing when weights were already downloaded and the program tried to redownload it. Now before every redownload the _remove_weights is called and even if the path is dir it will be removed.

jmatejcz: i added the note because it can seem unclear why we are deleting also dirs, but i think now it can be confusing, because the note suggest that the bug still exists

jmatejcz: i removed the note, leaving clearer comment 
image

maciejmajek (approved):

Issue #571, opened 2025-05-09
Killing ros2 processes


jmatejcz:
How should ros2 processes be killed? Can it be done safely from python?

jmatejcz:
It seems like some ros2 packages have just issue of not closing even after sending shutdown signal, for example moveit stays alive even after sending SIGINT signal or shutdown ros python api.

The https://github.com/RobotecAI/rai/commit/1e46ed7751346a5270e0bf24e5d4e69fb5679399 version escalates signal from SIGINT up to KILL for the process that called ros2  launch, but it  doesnt ensure that child processes will close

the newer solution -> https://github.com/RobotecAI/rai/pull/581, uses LaunchService which gives more control over the execution via ros2 python api but the problem stays the same.

If we need to 100% ensure that the nodes are closed, we probably would need to send pkill signals to every process

jmatejcz:
tried to apply approach from https://github.com/RobotecAI/rai/pull/445 to monitor processes, shutting down everything when any process is down, then starting all of them once again - That would be great for long running benchmarks, when something crashes, benchmark will continue after starting all processes again. Unfortunately i didn`t manage to force ros2 processes to behave properly so i leave it for now. maybe I'll come back to it when I have more time.

For now if something crashes mid running, the results are saved after every scenario so they wont be lost, but reruning rest of the scenarios will require doing it manually.

PR #581, opened 2025-05-14
refactor: running and killing ros2 processes

jmatejcz:
## Purpose
Current implementation of running and shutting down ros2 stack gives little control from python level.

Besides that i refactored simulation config logic as it was not best suited for current code.
the SimulationConfig contained both entities  which is required for scene setup and settings for simulation and robotic stack which are needed only when launching. This was not the cleanest by itself but also resulted in confusing configurations:
- Every single Scenario needed entites  but had also all the configuration for simulation 
- The ifs to run simulation were checked at every scenario

## Proposed Changes
Introduced Ros2LaunchManager which manages the launching and shutting down of ros nodes via LaunchService object from ros2 api.
Field `robotic_stack_command` removed from o3de_config as it was weird that user need to provide this. Now it is LaunchDescription is assigned to benchmarks. 

Refactored Simulation config that into:
- SimulationConfig that retained simulation configs
- SceneConfig - with entites

O3DExROS2SimulationConfig is now without `robotic_stack_command` field as it no longer needed

## Issues
https://github.com/RobotecAI/rai/issues/571

## Testing
1. Remove robotic_stack_command from your o3de_config.yaml
2. Run `python src/rai_bench/rai_bench/examples/test_models.py`
with settings:
```python
 models_name = ["qwen2.5:7b", "llama3.2"]
    vendors = ["ollama", "ollama"]
    benchmarks = ["manipulation_o3de"]
    extra_tool_calls = [5]
    repeats = 1
```
3. Check if nodes are properly closed after finish:
```bash
ros2 node list
```

### DIFF
--- a/src/rai_bench/rai_bench/examples/manipulation_o3de/main.py
+++ b/src/rai_bench/rai_bench/examples/manipulation_o3de/main.py
@@ -39,6 +39,7 @@ from rai_bench.utils import (
 )
 from rai_sim.o3de.o3de_bridge import (
     O3DEngineArmManipulationBridge,
+    O3DExROS2SimulationConfig,
 )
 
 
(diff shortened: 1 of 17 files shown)


jmatejcz:
I' m not sure where to put Ros2LaunchManager, for its `in rai_sim/launch_manager.py`

maciejmajek (approved):
Please rename Ros2LaunchManager to ROS2LaunchManager, other than that LGTM and GJ

PR #583, opened 2025-05-14
fix: remove print from test

rachwalk:
## Purpose

There shouldn't be print in test

## Proposed Changes

Remove print

## Issues

NA

## Testing

NA


maciejmajek (approved):
LGTM, well done.

PR #594, opened 2025-05-16
feat: respect timeout in service call ros2 api

### LINKED ISSUE #514: ROS2ServiceAPI.call_service timeout handling
**Is your feature request related to a problem? Please describe.**
Currently, the call_service method of the ROS2ServiceAPI does not respect the timeout parameter during the actual service call. The timeout is only applied during service discovery.
https://github.com/RobotecAI/rai/blob/0675ff6036bbd1450f34426f10d64f999ab91777/src/rai_core/rai/communication/ros2/api.py#L648-L674

**Describe the solution you'd like**
The timeout should apply to both the service client initialization and the service call phases. We should measure the time taken by the first phase and apply the remaining timeout to the service call.

maciejmajek:
## Purpose

Resolves https://github.com/RobotecAI/rai/issues/514

## Proposed Changes

Handling timeout sec

## Issues

Partially resolves #514 

## Testing

### DIFF
--- a/src/rai_core/rai/communication/ros2/connectors/service_mixin.py
+++ b/src/rai_core/rai/communication/ros2/connectors/service_mixin.py
@@ -34,7 +34,7 @@ class ROS2ServiceMixin:
         self,
         message: ROS2Message,
         target: str,
-        timeout_sec: float = 1.0,
+        timeout_sec: float = 5.0,
         *,
         msg_type: str,
         **kwargs: Any,
--- a/src/rai_core/rai/communication/ros2/api/service.py
+++ b/src/rai_core/rai/communication/ros2/api/service.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import uuid
 from typing import (
     Any,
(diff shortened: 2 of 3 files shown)


src/rai_core/rai/communication/ros2/connectors/service_mixin.py
```diff
@@ -34,7 +34,7 @@ def service_call(
         self,
         message: ROS2Message,
         target: str,
-        timeout_sec: float = 1.0,
+        timeout_sec: float = 5.0,
```
boczekbartek, line 37: For me 5 seconds for a service call is quite long, I would leave 1s

boczekbartek (approved):

src/rai_core/rai/communication/ros2/connectors/service_mixin.py
maciejmajek: I believe most services won't be able to respond in that time. 
Besides, this is the low level api which by default should provide safe parameters + most of the things using it should override the value to increase safety or optimize.

boczekbartek: ok, let's leave 5s then

boczekbartek:
@maciejmajek PR approved, it's fine to merge it now.

Issue #526, opened 2025-04-15
Action mocks


jmatejcz:
**Is your feature request related to a problem? Please describe.**
rai bench requires mocks of action interfaces, feedbacks and results to work properly

**Describe the solution you'd like**
Interfaces will be same as from ros2 cli - same as in topics and services.
Feedbacks and results also will be manually checked for every action type and mocked accordingly.

**Describe alternatives you've considered**

**Additional context**

What to do:
- [x] models
- [x] interfaces mocks
- [x] feedback mocks
- [x] results mocks
- [x] tool mocks

jmatejcz:
applied here: https://github.com/RobotecAI/rai/pull/595

Issue #462, opened 2025-03-19
Timeout for benchamark's scenario


jmatejcz:
**Is your feature request related to a problem? Please describe.**
Agent can be stuck on benchmark's single scenario for whatever reason

**Describe the solution you'd like**
Add timeout for running single scenario. Think about how to stop/reset simulation?

maciejmajek:
Hi @jmatejcz, did you manage to implement a solution?

jmatejcz:
applied here: https://github.com/RobotecAI/rai/pull/595

PR #595, opened 2025-05-19
feat: add missing logic rai_bench

jmatejcz:
## Purpose
Add couple missing elements:
- Action models, mocks to enable making tasks in tool calling agent benchmark that use actions
- timeouts for benchmarks ( single scenario or task )
- Langfuse integration for manipulation benchmark
- Validation tab in visualise script not working properly

Also Improve user experience from using the rai bench package as it feels a bit awkward from user perspective ( i concluded that when writing docs)

## Proposed Changes
- Fixed existing mocks of camera topics
- Added models and mocks of interfaces for messages used in navigation
- Added timeout for single scenario in Manipulation bench (210 sec) and in Tool Calling Agent (60 sec) 
- Added Langfuse for manipulation bench. Also moved the score tracing code to `results_processing/` dir

- Moved code related to predefined benchmarks from `examples/` to `<benchmark>/predefined` as tasks and scenarios defined by us should be part of package that user can import and use, rather than just example.  Example dir now contains really only code that imports out package and uses it.

- Introduced `test_models` function and benchmark configs that encapsulates all logic into just one function that can gather results from different models across different benchmarks. See `rai_bench/test_models.py` for code and `rai_bench/example/benchmarking_models.py` on how to use it.

- Results are now stored as: 1 run dir = 1 benchmark, this was required as now user can run same benchmark but with couple different set of params and we have to differentiate that.

- Visualise script now lets user chose couple runs and concatenate the results from them. Also they are sorted by date now.

- Fixed loading validation info  and now validator tab renders valid info

- Restructured  visualise code - seperate dir `results_processing/visualise/` , script divided in couple files so it is more readable.

- Added missing arguments to argparse.
## Issues
https://github.com/RobotecAI/rai/issues/526
https://github.com/RobotecAI/rai/issues/462
## Testing
1. testing multiple models in different configuration, visit the file, then run:
```bash
python src/rai_bench/rai_bench/examples/benchmarking_models.py 
```
2. Run 1 benchmark, experiment with new flags:
```bash
python src/rai_bench/rai_bench/examples/manipulation_o3de.py --model-name qwen2.5:7b --vendor ollama
```
```bash
python src/rai_bench/rai_bench/examples/tool_calling_agent.py --model-name qwen2.5:7b --vendor ollama
```
3. See the streamlit script, explore the UI
```bash
streamlit run  src/rai_bench/rai_bench/examples/visualise_streamlit.py
```


maciejmajek (approved):
LGTM

PR #597, opened 2025-05-21
fix: service call in agri demo

boczekbartek:
## Purpose

On ros2 humble, `examples/agriculture-demo.py` is not working due to the deadlock. Agent locks after the first service call.
image

Most likely it's due to the fact that the service is called in a timer. Such issues has 
been reported in the past in https://github.com/ros2/rclpy/issues/1016 
and https://github.com/ros2/rclpy/issues/1018.

## Proposed Changes

Execute agent in callback.

TODOs:
- [ ] service call api of ROS2Connector doesn't support callback.

## Testing

Follow steps from `docs/demos/agriculture.md`

### DIFF
--- a/examples/agriculture-demo.py
+++ b/examples/agriculture-demo.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import argparse
-from threading import Timer
 
 from langchain_core.messages import HumanMessage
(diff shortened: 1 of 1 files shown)


boczekbartek:
Closing. Replaced by: https://github.com/RobotecAI/rai/pull/599

PR #599, opened 2025-05-21
fix: agriculture demo deadlock on humble

boczekbartek:
## Purpose

On ros2 humble, `examples/agriculture-demo.py` is not working due to the deadlock. Agent locks after the first service call.
image

Most likely it's due to the fact that the service is called in a timer. Such issues has 
been reported in the past in https://github.com/ros2/rclpy/issues/1016 
and https://github.com/ros2/rclpy/issues/1018.

## Proposed Changes

Replace `threading.Timer` with a while loop.

## Testing
Follow `docs/demos/agriculture.md`

### DIFF
--- a/examples/agriculture-demo.py
+++ b/examples/agriculture-demo.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 import argparse
-from threading import Timer
+import threading
+import time
 
 from langchain_core.messages import HumanMessage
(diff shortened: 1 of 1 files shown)


maciejmajek (approved):
Apparently, threading.Timer calls function only once and is not causing any deadlocks. 
node.create_timer on the other hand, runs the functions periodically. 
The more you know...

Issue #464, opened 2025-03-21
Improve benchmark integration with langfuse/langsmith


MagdalenaKotynia:
## **Is your feature request related to a problem? Please describe.**
Related to #455.
Currently implemented integration with langfuse and langsmith could be improved.

### Proposed improvements:

**Tracking platform (CPU/GPU etc.) and what compute resource is currently used by the model:**

- **Purpose:** To reliably compare the latency between the models
- **Effort:** Low

**Adding task ID**

- **Purpose:** Sometimes I wanted to check tracking for a specific task and it was hard to search it. It can be useful for comparing models’ performance for a specific task
- **Effort:** Low

**Tracking commit hash**

- **Purpose:** to compare different versions of RAI, check performance after some changes in code, make sure that we compare something reliably
- **Effort:** Low

 **Tracking session ID**

- **Purpose:** To easily filter one benchmark run from all runs. 
- **Effort:** Low

**Introduce Error Codes**

- **Purpose:** To easily filter errrors/pass them to fine-tuning workflow
- **Effort:** Low

## **Additional context**
Errors included in `Comment` column could have error category IDs. (screen from Langfuse)
![Image](https://github.com/user-attachments/assets/a2465e93-97c8-476b-a982-b83378bb9a40)

Session column in langfuse could be used (in langsmith other mechanism could be used, I didn't notice explicit session id there)

![Image](https://github.com/user-attachments/assets/7f25d7e2-d30c-4d66-a5a0-4d7ed31b6ca9)

maciejmajek:
@MagdalenaKotynia what's the current timeline for this task? Anyone working on this?

MagdalenaKotynia:
@maciejmajek I haven't started working on this task. This enhancement proposal was created as a future enhancement after more priority work on the tool calling benchmark is done. I suggest to start working on it after the refactor of rai_bench in #517 is finished and merged. 
I edited the issue and added some other proposed improvements related to tracing.

maciejmajek:
I'm removing this issue from RAI 2.0, due to time constraints as well as low priority of the task

jmatejcz:
applied here https://github.com/RobotecAI/rai/pull/606

PR #606, opened 2025-05-29
feat: Dual agent + tracing tags

jmatejcz:
## Purpose
Enable usage of agents that compose of 2 llms. This can be usefull for testing the model which does not have tool calling capabilities. In such case we can use it in combination with llm that can make these calls.

Also this PR proposes changes to tracing to help debug and check results 
## Proposed Changes
- New way of creating agents -> `create_multimodal_to_tool_agent` function. It defines 2 llm like stated above, with addition of context bridge that removes images from context when passing it to 2nd model. This approach is needed when testing the 1st model on spatial reasoning tasks, as we ensure that the tool call is based only on  1st model's answer.

I know it's not a general solution for pairing llms but i think every task would require  slightly  different approach to defining  the langchain workflow. So when new workflow will be needed, just build something similar like in proposed function.

- new example file for showing how to run dual agent

- added tags to tracing for both tool-calling-agent and manipulation-o3de benchmarks
-  recursion limit in tool-calling-agent now scaling with number of langchain nodes, as more nodes = more recursion

## Issues
- https://github.com/RobotecAI/rai/issues/464
-   Links to relevant issues

## Testing
1. Run with tracing enabled 
```python 
python src/rai_bench/rai_bench/examples/dual_agent.py 
```
2. check the traces, if they are tagged clearly

### DIFF
--- a/src/rai_core/rai/agents/langchain/core/__init__.py
+++ b/src/rai_core/rai/agents/langchain/core/__init__.py
@@ -15,7 +15,6 @@
 from .conversational_agent import State as ConversationalAgentState
 from .conversational_agent import (
     create_conversational_agent,
-    create_multimodal_to_tool_agent,
 )
 from .react_agent import (
     ReActAgentState,
@@ -29,7 +28,6 @@ __all__ = [
     "ReActAgentState",
     "ToolRunner",
     "create_conversational_agent",
-    "create_multimodal_to_tool_agent",
     "create_react_runnable",
     "create_state_based_runnable",
 ]
--- a/src/rai_core/rai/agents/langchain/core/__init__.py
+++ b/src/rai_core/rai/agents/langchain/core/__init__.py
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 from .conversational_agent import State as ConversationalAgentState
-from .conversational_agent import create_conversational_agent
+from .conversational_agent import (
+    create_conversational_agent,
+    create_multimodal_to_tool_agent,
+)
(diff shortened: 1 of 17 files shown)


src/rai_core/rai/agents/langchain/core/conversational_agent.py
maciejmajek, line 1: Please move the new agent to the rai_bench package.

Processing images will be handled by rai_core in the future.

jmatejcz: what about the original `create_conversational_agent`? leave it there?

maciejmajek: Yes

jmatejcz: applied here: 
https://github.com/RobotecAI/rai/pull/606/commits/69035b3e1b652169fffeeba62142a15faeb71289

src/rai_core/rai/agents/langchain/core/__init__.py
maciejmajek, line 1: Please remove this change. otherwise lgtm

jmatejcz: applied here:
https://github.com/RobotecAI/rai/pull/606/commits/2512ec91349737c723dfa18cdbbd412f87ff256f

maciejmajek (approved):
LGTM

PR #615, opened 2025-06-02
fix: Update poetry.lock

knicked:
## Purpose

-   `sam2` dependency in current `poetry.lock` file points to a commit that doesn't exist, leading to this error when doing `poetry install --all-groups` on a fresh environment: 
```
- Installing sam2 (1.0 335de4d): Failed

    | Failed to clone https://github.com/RobotecAI/Grounded-SAM-2 at '335de4d0d8ef1960865f05d7da3a266b99d895a8', verify ref exists on remote.
    | 
    | Note: This error arises from interacting with the specified vcs source and is likely not a Poetry issue.
    | This issue could be caused by any of the following;
    | 
    | - there are network issues in this environment
    | - the revision (335de4d0d8ef1960865f05d7da3a266b99d895a8) you have specified
    |     - was misspelled
    |     - is invalid (must be a sha or symref)
    |     - is not present on remote
```

## Proposed Changes

Updated `poetry.lock` file by running `poetry lock`

## Issues

- N/A

## Testing

-  `poetry install --all-groups` was ran successfully.


maciejmajek (approved):

maciejmajek (approved):
Replicated, fix works

Issue #576, opened 2025-05-12
define final tasks in tool calling benchmark


jmatejcz:
**Is your feature request related to a problem? Please describe.**
For now we have 30 tasks in Tool calling agent benchmark from categories:
- basic (get image from camera topic etc.)
- navigation ( navigate somewhere type tasks )
- manipulation ( grab / drop something )
- spatial reasoning ( answer question about an image )
- custom interfaces ( forming and sending the messages of custom interface )

**Describe the solution you'd like**
We need to establish what tasks we want and how many.
Please check out current tasks. 
If you have any suggestion, leave a comment

**Describe alternatives you've considered**

**Additional context**

jmatejcz:
So my main questions:
- Should there be more categories? 
- What categories should have more tasks than others? 
IMO:
 basic - so getting something from a right topic are pretty similar so no need for large amount?
 custom_interfaces - quite similar but we can change what need to be send so there could be more 
 manipulation/ navigation/ spatial_reasoning could be created in large numbers but it would require some specialise verification process so it can take some time

- How should we format the prompts, currently we have a mix of very detailed prompts
like in in custom_interfaces tasks:
```python
f"You need to publish a message to the topic '{self.topic}' with the text value: '{self.text}'.\n"
"Before publishing, follow these steps:\n"
"1. Use the tool to retrieve the available ROS2 topics and their message types.\n"
f"2. Find the message type for the topic '{self.topic}'.\n"
"3. Retrieve the full message interface definition for that type.\n"
"4. Construct the message filling only the fields you are instructed to. Rest of the fields will have default values.\n"
f"5. Publish the message to '{self.topic}' using the correct message type and interface.\n"
```
but on the other hand there are prompts like in navigation:
```python 
"Move 2 meters to the front."
```

My proposal to this question is separate tasks for both types of prompts and mark them with different complexity (we have easy, medium, hard), but i m not sure as this was designed to represent how hard the task itself is. 

4. System prompts - how detailed they should be, we have like this for manipulation, with no examples or anything really:
```python 
You are a robotic arm with interfaces to detect and manipulate objects.
Here are the coordinates information:
x - front to back (positive is forward)
y - left to right (positive is right)
z - up to down (positive is up).
```
but this monster for navigation:
```python
"""You are an autonomous robot connected to ros2 environment. Your main goal is to fulfill the user's requests.
    Do not make assumptions about the environment you are currently in.
    You can use ros2 topics, services and actions to operate.

[...]

PR #638, opened 2025-06-20
refactor: remove navigation tasks

jmatejcz:
## Purpose
Navigational tasks are hard to validate or even impossible to validate based solely on tool calls. Unlike manipulation, where we have a tool Manipulator move to, here the agent is operating on nav2 actions which makes it harder to validate. Additionally in navigation there can be nearly infinite number of ways to drive somewhere. For example to move 20cm forward i can call move 10cm forward action 2 times and it would result in the same movement, therefor we can't say that the 2nd approach is wrong.

## Proposed Changes
Based on that i decided to remove navigation tasks from tool calling benchmark. If you would want to have benchmark for navigation in the future, i think it makes much more sense to validate it via end result in simulation environment.
## Issues
Partially https://github.com/RobotecAI/rai/issues/576


boczekbartek (approved):
@jmatejcz 
I agree that testing navigation tasks is not feasible without the simulation. 
I am fine with removing the code, but how about creating and mentioning this PR. I think the code can be reused in simulation benchmarks if they will be implemented in the future.

jmatejcz:
> @jmatejcz I agree that testing navigation tasks is not feasible without the simulation. I am fine with removing the code, but how about creating and mentioning this PR. I think the code can be reused in simulation benchmarks if they will be implemented in the future.

yea sure, i will add issue for future and mention this PR:https://github.com/RobotecAI/rai/issues/657
i leave it as question label as it is not planned to be done yet

Issue #623, opened 2025-06-06
Tool calling agent args of calls validation


jmatejcz:
**Is your feature request related to a problem? Please describe.**
A validation of arguments send as message, service args or actions args is needed
Now we have only validation for custom interfaces which are validated against pydantic models to ensure that fields are proper. But we also need to validate all other interfaces .

Also current implementation of  checking function like `_check_service_tool_call_field` which is supposed to check if the content of message is valid, does not properly validate list that are appearing in arguments

**Describe the solution you'd like**
Validation should compose of 2 paths, one for `common` interfaces which will be validated against imported types, second for custom interfaces. 

**Describe alternatives you've considered**

**Additional context**

jmatejcz:
added in PR -> https://github.com/RobotecAI/rai/pull/644
commit to main -> https://github.com/RobotecAI/rai/commit/baace12eb70bb761a3fdf1aa81dcf3e43aaa9d59

Issue #622, opened 2025-06-06
Tool calling benchmark more validators


jmatejcz:
**Is your feature request related to a problem? Please describe.**
For now we have 2 type of validators:
- ordered, which checks if all subtasks execute in certain order
- not ordered, which checks if all subtasks execute in any order

There is a need for other type that will pass when only 1 of given subtasks passes. This will be useful for tasks that can be done in couple ways.

Second kind of problem is when there is multiple paths of completing the task. We won't bother with tasks that have large number of paths like navigation, but we can consider tasks with 2 or 3 possible ways. For example swapping objects.

**Describe the solution you'd like**
OptionalValidator class and MuliPathValidator class

**Describe alternatives you've considered**

**Additional context**

jmatejcz:
MuliPathValidator which would take couple lists of subtasks is not really coherent with current implementation of Validator, so i think i won't implement that.  So our tasks should have no more than 1 path of completion and the prompt should enforce this path

jmatejcz:
added here: https://github.com/RobotecAI/rai/pull/644
commit to main -> https://github.com/RobotecAI/rai/commit/baace12eb70bb761a3fdf1aa81dcf3e43aaa9d59

PR #644, opened 2025-07-01
feat: basic tasks extension

jmatejcz:
## Purpose
Make new basic tasks for tool calling benchmark, test and adjust task prompts and system prompt
For now basic tasks revolve around receiving messages from topics, which is quite easy. we need more medium and hard tasks.

## Proposed Changes
- To make some harder tasks i included service calls. To validate these calls we nned to check is provided service type exists and if args match this type.To do this I implemented `ServiceValidator`. It combines the validation of custom interfaces, that was already present and  done with pydantic models, with validation of ROS2 services present in installed packages.

- Introduced `_check_nested_fields` which resolved issue with checking nested fields that did not work when  field was list.
Now you can use int as idx when passing path:
```python
set_robot_state_params_subtask = CheckServiceFieldsToolCallSubTask(
    expected_tool_name="call_ros2_service",
    expected_service="/robot_state_publisher/set_parameters",
    expected_service_type="rcl_interfaces/srv/SetParameters",
    expected_fields={
        "parameters.0.name": "publish_frequency",
        "parameters.0.value.type": "3",
        "parameters.0.value.double_value": 30.0,
    },
)
``` 
- Added new type of validator - Optional Validator - it passes when 1 of given subtasks passes. Useful when same thing can be accomplished in 2 different tool calls.  
- Recurssion limit calculated differently so it scales better with number of extra and optional tool calls 
- Added a bunch of mocked interfaces to match all types written in available topics
- Added 7 new tasks of medium and hard complexity, that makes total of 14 tasks.
- Some of new tasks are parametrizable so they can be created with different items and values in prompts.
- Currently there are 16 predefined, importable basic tasks

- Added tests for predefined tasks to make sure they pass and fail under certain tool calls. 
- Added tests for ServiceValidator
- Added tests with nested lists to subtask tests and adjusted typing
## Issues
partially https://github.com/RobotecAI/rai/issues/576
https://github.com/RobotecAI/rai/issues/623
https://github.com/RobotecAI/rai/issues/622

## Testing
1. Mark only basic tasks, then run;
```bash
python src/rai_bench/rai_bench/examples/benchmarking_models.py 
```
2. View results, check if everything works 
3. Run tests
```bash
pytest tests/rai_bench/tool_calling_agent
```
4. Check the task prompts in `rai_bench/tool_calling_agent/tasks/basic.py` if they make sense to you
5. Check the subtasks in `rai_bench/tool_calling_agent/predefined/tasks_tasks.py` if they make sense to you

## Summary by CodeRabbit

- **New Features**
  - Added advanced support for ROS2 service discovery, parameter management, entity spawning, and AI vision pipeline configuration in predefined tasks.
  - Introduced new validators, including an optional validator, enhancing task validation flexibility.
  - Expanded the set of available ROS2 interface definitions for more comprehensive simulation and testing.

- **Refactor**
  - Streamlined and centralized nested field validation logic for subtasks, reducing code duplication.
  - Reorganized and extended predefined tasks to focus more on service-based and parameter-centric operations.

- **Bug Fixes**
  - Improved error handling and validation for ROS2 service calls and argument checking.

- **Tests**
  - Added extensive new test suites covering service validation, predefined tasks, subtasks, and validator logic to ensure robustness and correctness.

### DIFF
--- a/src/rai_bench/rai_bench/tool_calling_agent/benchmark.py
+++ b/src/rai_bench/rai_bench/tool_calling_agent/benchmark.py
@@ -87,7 +87,7 @@ class ToolCallingAgentBenchmark(BaseBenchmark):
         )
         callbacks = self.score_tracing_handler.get_callbacks()
         run_id = uuid.uuid4()
-        # NOTE (jmatejcz) reccustion limit calculated as all_nodes_num -> one pass though whole node
+        # NOTE (jmatejcz) recursion limit calculated as all_nodes_num -> one pass though whole node
         # plus (task.max_tool_calls_number-1 because the first pass is already added in)
         # times number of nodes - 2 because we dont cout start and end node
         # this can be to much for larger graphs that dont use all nodes on extra calls
--- a/src/rai_bench/rai_bench/tool_calling_agent/benchmark.py
+++ b/src/rai_bench/rai_bench/tool_calling_agent/benchmark.py
@@ -87,6 +87,14 @@ class ToolCallingAgentBenchmark(BaseBenchmark):
         )
         callbacks = self.score_tracing_handler.get_callbacks()
         run_id = uuid.uuid4()
+        # NOTE (jmatejcz) reccustion limit calculated as all_nodes_num -> one pass though whole node
+        # plus (task.max_tool_calls_number-1 because the first pass is already added in)
+        # times number of nodes - 2 because we dont cout start and end node
+        # this can be to much for larger graphs that dont use all nodes on extra calls
+        # in such ase adjust this value
+        recurssion_limit = len(agent.get_graph().nodes) + (
+            task.max_tool_calls_number - 1
+        ) * (len(agent.get_graph().nodes) - 2)
         config: RunnableConfig = {
             "run_id": run_id,
             "callbacks": callbacks,
@@ -97,9 +105,9 @@ class ToolCallingAgentBenchmark(BaseBenchmark):
                 f"task-complexity:{task.complexity}",
                 f"extra-tool-calls:{task.extra_tool_calls}",
             ],
-            "recursion_limit": len(agent.get_graph().nodes)
-            * task.max_tool_calls_number,
+            "recursion_limit": recurssion_limit,
         }
+        self.logger.debug(f"recurssion limit: {recurssion_limit}")
 
         ts = time.perf_counter()
         messages: List[BaseMessage] = []
--- a/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
+++ b/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
@@ -269,14 +269,40 @@ class MockGetROS2MessageInterfaceTool(GetROS2MessageInterfaceTool):
 
 
 class ServiceValidator:
-    """Hybrid validator that uses ROS 2 native types when available, falls back to Pydantic"""
+    """
+    Hybrid validator that uses ROS 2 native types when available,
+    falls back to Pydantic models of custom interfaces
+    """
 
     def __init__(self, custom_models: Dict[str, Type[BaseModel]]):
         self.custom_models = custom_models
         self.ros2_services_cache: Dict[str, Any] = {}
 
     def get_ros2_service_class(self, service_type: str):
-        """Dynamically import ROS 2 service class"""
+        """
+        Dynamically import ROS 2 service class.
+
+        Parameters
+        ----------
+        service_type : str
+            The ROS 2 service type in format "package_name/srv/ServiceName".
+
+        Returns
+        -------
+        class
+            The dynamically imported ROS 2 service class.
+
+        Raises
+        ------
+        ValueError
+            If the service_type format is invalid (not in format
+            "package_name/srv/ServiceName").
+
+        Notes
+        -----
+        Results are cached in ros2_services_cache to avoid repeated imports
+        of the same service type.
+        """
         if service_type in self.ros2_services_cache:
             return self.ros2_services_cache[service_type]
 
@@ -294,12 +320,18 @@ class ServiceValidator:
         return service_class
 
     def validate_with_ros2(self, service_type: str, args: Dict[str, Any]):
-        """Validate using native ROS 2 service definition
+        """Validate using installed ROS2 packages services definition
+
+        Parameters
+        ----------
+        service_type : str
+        args : Dict[str, Any]
+            Dictionary of arguments to validate against the service definition.
 
         Raises
         ------
         TypeError
-            When service type
+            When service type does not exist in ROS2 installed packages
         """
         service_class = self.get_ros2_service_class(service_type)
         if not service_class:
@@ -311,7 +343,21 @@ class ServiceValidator:
         set_message_fields(request, args_to_validate)
 
     def validate_with_custom(self, service_type: str, args: Dict[str, Any]):
-        """Validate using Pydantic model"""
+        """
+        Validate using Pydantic model of custom messages.
+
+        Parameters
+        ----------
+        service_type : str
+        args : Dict[str, Any]
+            Dictionary of arguments to validate against the Pydantic model.
+
+        Raises
+        ------
+        ValueError
+            If service_type is not found in custom_models or if Pydantic
+            validation fails.
+        """
         if service_type not in self.custom_models:
             raise ValueError(f"Service type: {service_type} is invalid custom type")
 
@@ -322,7 +368,15 @@ class ServiceValidator:
             raise ValueError(f"Pydantic validation failed: {e}")
 
     def validate(self, service_type: str, args: Dict[str, Any]):
-        """Try ROS 2 validation first, fall back to Pydantic"""
+        """
+        Try ROS 2 validation first, fall back to Pydantic models.
+
+        Parameters
+        ----------
+        service_type : str
+        args : Dict[str, Any]
+            Dictionary of arguments to validate.
+        """
         if service_type in self.custom_models:
             self.validate_with_custom(service_type, args)
         else:
@@ -338,6 +392,7 @@ class MockCallROS2ServiceTool(CallROS2ServiceTool):
     @computed_field
     @property
     def models_validator(self) -> ServiceValidator:
+        """computed field for instancinating ServiceValidator with available service models"""
         return ServiceValidator(self.available_service_models)
 
     def _run(
--- a/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
+++ b/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
@@ -12,15 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
+import importlib
 import uuid
 from threading import Lock
-from typing import Any, Dict, List, Tuple, Type
+from typing import Any, Dict, List, Optional, Tuple, Type
 from unittest.mock import MagicMock
 
 import numpy as np
 import numpy.typing as npt
 from langchain_core.tools import BaseTool
-from pydantic import BaseModel, ValidationError
+from pydantic import BaseModel, ValidationError, computed_field
 from rai.communication.ros2.connectors import ROS2Connector
 from rai.communication.ros2.messages import ROS2Message
 from rai.messages import MultimodalArtifact, preprocess_image
@@ -48,6 +50,7 @@ from rai_open_set_vision.tools import (
     GetDistanceToObjectsTool,
     GetGrabbingPointTool,
 )
+from rosidl_runtime_py import set_message_fields
 
 
 class MockGetROS2TopicsNamesAndTypesTool(GetROS2TopicsNamesAndTypesTool):
@@ -265,21 +268,88 @@ class MockGetROS2MessageInterfaceTool(GetROS2MessageInterfaceTool):
             raise ImportError(f"Module {msg_type} not found.")
 
 
+class ServiceValidator:
+    """Hybrid validator that uses ROS 2 native types when available, falls back to Pydantic"""
+
+    def __init__(self, custom_models: Dict[str, Type[BaseModel]]):
+        self.custom_models = custom_models
+        self.ros2_services_cache: Dict[str, Any] = {}
+
+    def get_ros2_service_class(self, service_type: str):
+        """Dynamically import ROS 2 service class"""
+        if service_type in self.ros2_services_cache:
+            return self.ros2_services_cache[service_type]
+
+        # Parse service type: "package_name/srv/ServiceName"
+        parts = service_type.split("/")
+        if len(parts) != 3 or parts[1] != "srv":
+            raise ValueError(f"Service type: {service_type} is invalid")
+
+        package_name, _, service_name = parts
+
+        module = importlib.import_module(f"{package_name}.srv")
+        service_class = getattr(module, service_name)
+
+        self.ros2_services_cache[service_type] = service_class
+        return service_class
+
+    def validate_with_ros2(self, service_type: str, args: Dict[str, Any]):
+        """Validate using native ROS 2 service definition
+
+        Raises
+        ------
+        TypeError
+            When service type
+        """
+        service_class = self.get_ros2_service_class(service_type)
+        if not service_class:
+            raise TypeError(f"Service type: {service_type} does not exist.")
+
+        request = service_class.Request()
+        # set message fields converts them to object so we need deepcopy to avoid it
+        args_to_validate = copy.deepcopy(args)
+        set_message_fields(request, args_to_validate)
+
+    def validate_with_custom(self, service_type: str, args: Dict[str, Any]):
+        """Validate using Pydantic model"""
+        if service_type not in self.custom_models:
+            raise ValueError(f"Service type: {service_type} is invalid custom type")
+
+        model = self.custom_models[service_type]
+        try:
+            model.model_validate(args)
+        except ValidationError as e:
+            raise ValueError(f"Pydantic validation failed: {e}")
+
+    def validate(self, service_type: str, args: Dict[str, Any]):
+        """Try ROS 2 validation first, fall back to Pydantic"""
+        if service_type in self.custom_models:
+            self.validate_with_custom(service_type, args)
+        else:
+            return self.validate_with_ros2(service_type, args)
+
+
 class MockCallROS2ServiceTool(CallROS2ServiceTool):
     connector: ROS2Connector = MagicMock(spec=ROS2Connector)
     available_services: List[str]
     available_service_types: List[str]
     available_service_models: Dict[str, Type[BaseModel]]
 
+    @computed_field
+    @property
+    def models_validator(self) -> ServiceValidator:
+        return ServiceValidator(self.available_service_models)
+
     def _run(
         self,
         service_name: str,
         service_type: str,
-        service_args: Dict[str, Any],
+        service_args: Optional[Dict[str, Any]] = None,
+        timeout_sec: float = 1.0,
     ) -> str:
         if service_name not in self.available_services:
             raise ValueError(
-                f"Service {service_name} is not available within 1.0 seconds. Check if the service exists."
+                f"Service {service_name} is not available within {timeout_sec} seconds. Check if the service exists."
             )
         if service_type not in self.available_service_types:
             raise TypeError(
@@ -287,12 +357,10 @@ class MockCallROS2ServiceTool(CallROS2ServiceTool):
                     self.available_service_types, service_type
                 )
             )
-        if service_type in self.available_service_models:
-            model = self.available_service_models[service_type]
-            try:
-                model.model_validate(service_args)
-            except ValidationError as e:
-                raise ValueError(f"Failed to populate fields: {e}")
+        if not service_args:
+            service_args = {}
+        try:
+            self.models_validator.validate(service_type, service_args)
             response = ROS2Message(payload={"response": "success"})
             return str(
                 {
@@ -300,10 +368,8 @@ class MockCallROS2ServiceTool(CallROS2ServiceTool):
                     "metadata": response.metadata,
                 }
             )
-        else:
-            raise KeyError(
-                f"Model for service type {service_type} not included in models"
-            )
+        except ValueError as e:
+            raise ValueError(f"Failed to populate fields: {e}")
 
 
 class MockGetROS2ServicesNamesAndTypesTool(GetROS2ServicesNamesAndTypesTool):
--- a/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
+++ b/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
@@ -365,7 +365,7 @@ class ServiceValidator:
         try:
             model.model_validate(args)
         except ValidationError as e:
-            raise ValueError(f"Pydantic validation failed: {e}")
+            raise ValueError(f"Pydantic validation failed: {e}") from e
 
     def validate(self, service_type: str, args: Dict[str, Any]):
         """
@@ -380,7 +380,7 @@ class ServiceValidator:
         if service_type in self.custom_models:
             self.validate_with_custom(service_type, args)
         else:
-            return self.validate_with_ros2(service_type, args)
+            self.validate_with_ros2(service_type, args)
 
 
 class MockCallROS2ServiceTool(CallROS2ServiceTool):
@@ -402,6 +402,18 @@ class MockCallROS2ServiceTool(CallROS2ServiceTool):
         service_args: Optional[Dict[str, Any]] = None,
         timeout_sec: float = 1.0,
     ) -> str:
+        """
+        Execute the mocked ROS2 service call with validation of service type and its args.
+
+        Parameters
+        ----------
+        service_name : str
+            Name of the service to call
+        service_type : str
+            Type of the service
+        service_args : Optional[Dict[str, Any]], optional
+            Arguments for the service call, by default None
+        """
         if service_name not in self.available_services:
             raise ValueError(
                 f"Service {service_name} is not available within {timeout_sec} seconds. Check if the service exists."
--- a/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
+++ b/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
@@ -270,8 +270,10 @@ class MockGetROS2MessageInterfaceTool(GetROS2MessageInterfaceTool):
 
 class ServiceValidator:
     """
-    Hybrid validator that uses ROS 2 native types when available,
-    falls back to Pydantic models of custom interfaces
+    Validator that is responsible for checking if gicen service type exists
+    and if it is used correctly.
+    Validator uses ROS 2 native types when available,
+    falls back to Pydantic models of custom interfaces when not.
     """
 
     def __init__(self, custom_models: Dict[str, Type[BaseModel]]):
--- a/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
+++ b/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import copy
-import importlib
 import uuid
 from threading import Lock
 from typing import Any, Dict, List, Optional, Tuple, Type
@@ -23,6 +22,7 @@ import numpy as np
 import numpy.typing as npt
 from langchain_core.tools import BaseTool
 from pydantic import BaseModel, ValidationError, computed_field
+from rai.communication.ros2.api.conversion import import_message_from_str
 from rai.communication.ros2.connectors import ROS2Connector
 from rai.communication.ros2.messages import ROS2Message
 from rai.messages import MultimodalArtifact, preprocess_image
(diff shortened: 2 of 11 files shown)


src/rai_bench/rai_bench/tool_calling_agent/benchmark.py
```diff
@@ -87,6 +87,14 @@ def run_next(self, agent: CompiledStateGraph, experiment_id: uuid.UUID) -> None:
         )
         callbacks = self.score_tracing_handler.get_callbacks()
         run_id = uuid.uuid4()
+        # NOTE (jmatejcz) reccustion limit calculated as all_nodes_num -> one pass though whole node
```
boczekbartek:
```suggestion
        # NOTE (jmatejcz) recursion limit calculated as all_nodes_num -> one pass though whole node
```

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
```diff
@@ -265,45 +268,177 @@ def _run(self, msg_type: str) -> str:
             raise ImportError(f"Module {msg_type} not found.")
 
 
+class ServiceValidator:
+    """
+    Validator that is responsible for checking if gicen service type exists
```
boczekbartek:
```suggestion
    Validator that is responsible for checking if given service type exists
```

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
```diff
@@ -265,45 +268,177 @@ def _run(self, msg_type: str) -> str:
...
+    and if it is used correctly.
+    Validator uses ROS 2 native types when available,
+    falls back to Pydantic models of custom interfaces when not.
+    """
+
+    def __init__(self, custom_models: Dict[str, Type[BaseModel]]):
+        self.custom_models = custom_models
+        self.ros2_services_cache: Dict[str, Any] = {}
+
+    def get_ros2_service_class(self, service_type: str):
```
boczekbartek: return typehint is missing

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
```diff
@@ -265,45 +268,177 @@ def _run(self, msg_type: str) -> str:
...
+        parts = service_type.split("/")
+        if len(parts) != 3 or parts[1] != "srv":
+            raise ValueError(f"Service type: {service_type} is invalid")
+
+        package_name, _, service_name = parts
+
+        module = importlib.import_module(f"{package_name}.srv")
+        service_class = getattr(module, service_name)
+
+        self.ros2_services_cache[service_type] = service_class
```
boczekbartek: consider using:
https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/src/rai_core/rai/communication/ros2/api/conversion.py#L48

boczekbartek: consider using:
https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/src/rai_core/rai/communication/ros2/api/conversion.py#L48

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
```diff
@@ -265,45 +268,177 @@ def _run(self, msg_type: str) -> str:
...
+        Dynamically import ROS 2 service class.
+
+        Parameters
+        ----------
+        service_type : str
+            The ROS 2 service type in format "package_name/srv/ServiceName".
+
+        Returns
+        -------
+        class
```
boczekbartek: Looking at the further code I think None can be returned as well

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
```diff
@@ -265,45 +268,177 @@ def _run(self, msg_type: str) -> str:
...
+
+        package_name, _, service_name = parts
+
+        module = importlib.import_module(f"{package_name}.srv")
+        service_class = getattr(module, service_name)
+
+        self.ros2_services_cache[service_type] = service_class
+        return service_class
+
+    def validate_with_ros2(self, service_type: str, args: Dict[str, Any]):
```
boczekbartek, line 283: what happens when the validation fails?

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
```diff
@@ -265,45 +268,177 @@ def _run(self, msg_type: str) -> str:
...
+        if service_type not in self.custom_models:
+            raise ValueError(f"Service type: {service_type} is invalid custom type")
+
+        model = self.custom_models[service_type]
+        try:
+            model.model_validate(args)
+        except ValidationError as e:
+            raise ValueError(f"Pydantic validation failed: {e}") from e
+
+    def validate(self, service_type: str, args: Dict[str, Any]):
```
boczekbartek, line 331: what happens when the validation fails?
In case of `validate_with_custom` a ValueError is risen. 
How about `validate_with_ros2`?

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
```diff
@@ -265,45 +268,177 @@ def _run(self, msg_type: str) -> str:
...
+            Name of the service to call
+        service_type : str
+            Type of the service
+        service_args : Optional[Dict[str, Any]], optional
+            Arguments for the service call, by default None
+        """
         if service_name not in self.available_services:
             raise ValueError(
-                f"Service {service_name} is not available within 1.0 seconds. Check if the service exists."
+                f"Service {service_name} is not available within {timeout_sec} seconds. Check if the service exists."
```
boczekbartek, line 380: is this required in mock if `available_services` are constant?

src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py
```diff
@@ -20,164 +20,375 @@
...
+    expected_service_type=SET_PARAMETERS_ATOMICALLY_TYPE,
+    expected_fields={
+        "parameters.0.name": "fps",
+        "parameters.0.value.type": 2,
+        "parameters.0.value.integer_value": FPS_2,
+    },
+)
+
+
+delete_entity_subtask1 = CheckServiceFieldsToolCallSubTask(
```
boczekbartek: do we need all these tasks by name?
Wouldn't it be possible to generate these tasks with some factory function with much less code?

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
```diff
@@ -180,40 +219,243 @@ def get_prompt(self) -> str:
...
 
 
-class CheckRobotHealthTask(BasicTask):
+#### calling services ####
+class GetROS2ServicesTask(BasicTask):
+    complexity = "easy"
+
+    @property
+    def optional_tool_calls_number(self) -> int:
+        return 0
```
boczekbartek, line 300:
```suggestion
    optional_tool_calls_number = 0
```

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
```diff
@@ -180,40 +219,243 @@ def get_prompt(self) -> str:
...
+        task_args: TaskArgs,
+        logger: logging.Logger | None = None,
+    ) -> None:
+        super().__init__(validators, task_args, logger)
+        self.parameter = parameter
+
+    @property
+    def optional_tool_calls_number(self) -> int:
+        # list services and get interfaces
+        return 2
```
boczekbartek, line 357:
```suggestion
    optional_tool_calls_number = 2
```

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
```diff
@@ -180,40 +219,243 @@ def get_prompt(self) -> str:
...
+        value: float,
+        validators: List[Validator],
+        task_args: TaskArgs,
+        logger: logging.Logger | None = None,
+    ) -> None:
+        super().__init__(validators, task_args, logger)
+        self.value = value
+
+    @property
+    def optional_tool_calls_number(self) -> int:
```
boczekbartek, line 419: same as above

src/rai_bench/rai_bench/tool_calling_agent/validators.py
```diff
@@ -144,3 +144,54 @@ def validate(self, tool_calls: List[ToolCall]) -> Tuple[bool, List[ToolCall]]:
...
+            raise ValueError("Validator must have at least 1 subtask.")
+
+    @property
+    def type(self) -> str:
+        return "optional"
+
+    @property
+    def required_calls(self) -> int:
+        # For optional validator, we only need 1 call to pass any subtask
+        return 1
```
boczekbartek, line 168: covert to simple properties, since no logic is required for these properties

boczekbartek, line 168: covert to simple properties, since no logic is required for these properties

src/rai_bench/rai_bench/tool_calling_agent/validators.py
```diff
@@ -144,3 +144,54 @@ def validate(self, tool_calls: List[ToolCall]) -> Tuple[bool, List[ToolCall]]:
         if len(tool_calls) > self.required_calls:
             self.extra_calls_used = len(tool_calls) - self.required_calls
         return False, []
+
+
+class OptionalValidator(Validator):
```
boczekbartek: How about `OneFromManyValidator`?

boczekbartek: How about `OneFromManyValidator`?

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
```diff
@@ -180,40 +219,243 @@ def get_prompt(self) -> str:
...
-                "You can explore available diagnostic topics and gather information "
-                "about robot health, joint states, and system logs."
+                f"{self.get_base_prompt()} using parameter service. "
+                "You can list available services to find the appropriate service, "
+                f"check its type's interface and set the publish_frequency parameter to {self.value}."
             )
 
 
-class AssessSensorDataQualityTask(BasicTask):
+class CheckSpawnableEntitiesTask(BasicTask):
```
boczekbartek, line 437: This task is not very generic and has only 1 connected subtask/validator options.
How about just setting them in `__init__`, without parametrization? It would be easier to find parameters

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
```diff
@@ -180,40 +219,243 @@ def get_prompt(self) -> str:
...
+        logger: logging.Logger | None = None,
+    ) -> None:
+        super().__init__(validators, task_args, logger)
+        self.sam_confidence_threshold = sam_confidence_threshold
+        self.dino_confidence_threshold = dino_confidence_threshold
+        self.fps = fps
+
+    @property
+    def optional_tool_calls_number(self) -> int:
+        return 2  # list services, get interface
```
boczekbartek, line 605: would it be important to verify if exactly these 2 services were called as optional?

src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py
```diff
@@ -20,164 +20,375 @@
...
+    expected_service_type=SET_PARAMETERS_ATOMICALLY_TYPE,
+    expected_fields={
+        "parameters.0.name": "publish_frequency",
+        "parameters.0.value.type": "3",
+        "parameters.0.value.double_value": DEFAULT_PUBLISH_FREQUENCY,
+    },
+)
+
+
+set_grounded_sam_subtask_1 = CheckServiceFieldsToolCallSubTask(
```
boczekbartek: do we need 2 tasks just with different confidence threshold?
`set_grounded_sam_subtask_1` is the same but with different float

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
```diff
@@ -180,40 +219,243 @@ def get_prompt(self) -> str:
...
-                f"{self.get_base_prompt()} by examining system diagnostics and monitoring data. "
-                "You can explore available diagnostic topics and gather information "
-                "about robot health, joint states, and system logs."
+                f"{self.get_base_prompt()} using parameter service. "
+                "You can list available services to find the appropriate service, "
+                f"check its type's interface and set the publish_frequency parameter to {self.value}."
             )
 
 
-class AssessSensorDataQualityTask(BasicTask):
```
boczekbartek, line 204: Why is this task removed?

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
boczekbartek: Looking at tests it will raise `AttributeError` or `TypeError`.

https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/tests/rai_bench/tool_calling_agent/test_mock_tools.py#L117-L132

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
boczekbartek: [Looking at tests](https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/tests/rai_bench/tool_calling_agent/test_mock_tools.py#L131) it seems like `AttributeError` or `TypeError` will be raised, which seem not to be handled here:

https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py#L431-L441

boczekbartek (changes requested):
@jmatejcz Good job 🦾 I left some comments - could you have a look?

boczekbartek (changes requested):
@jmatejcz Good job 🦾 I left some comments - could you have a look?

src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py
jmatejcz: yes i know, do you think it's redundant to have 2 such tasks?

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
jmatejcz: as disscussed in person we leave as it is

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
jmatejcz: as disscussed in person we leave as it is

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
jmatejcz: not really, i think calling only 1 of them is also a valid

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
jmatejcz: it emulates the real response, will agent call non existing topic, that will be the anwser

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
jmatejcz: yes, the error will be risen when calling tool and any Error risen in Tool with be caught by our ToolRunner and returned to the agent

src/rai_bench/rai_bench/tool_calling_agent/benchmark.py
jmatejcz: applied here: https://github.com/RobotecAI/rai/pull/644/commits/cfdf1ee6587156df1fee083ae03dddf603c93e35

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
jmatejcz: applied here:
https://github.com/RobotecAI/rai/pull/644/commits/cfdf1ee6587156df1fee083ae03dddf603c93e35

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
jmatejcz: same as above

src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py
jmatejcz: but we need to have these expected fields and values either way, do you have some solution in mind?

If factories used there will be indeed less lines, but the variables stays i think

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
jmatejcz: as discussed in person, it stays

src/rai_bench/rai_bench/tool_calling_agent/validators.py
jmatejcz: as discussed in person, it stays

jmatejcz: as discussed in person, it stays

src/rai_bench/rai_bench/tool_calling_agent/validators.py
jmatejcz: duplicated

jmatejcz: yes, that more meaningful, applied here: https://github.com/RobotecAI/rai/pull/644/commits/cfdf1ee6587156df1fee083ae03dddf603c93e35

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
jmatejcz: I m not sure what do you mean, can you explain?

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
jmatejcz: this func removed here: https://github.com/RobotecAI/rai/pull/644/commits/b2dfc6914631f76c2942895a77b699e6c98eabd3

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
jmatejcz: applied here: https://github.com/RobotecAI/rai/pull/644/commits/b2dfc6914631f76c2942895a77b699e6c98eabd3

If i understand correctly, this do exactly the same as my function did (all unittests passes also), so i just removed my fucntion entirely for this. Correct me if I'm wrong pls

jmatejcz: duplicated

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
jmatejcz: func removed here: https://github.com/RobotecAI/rai/pull/644/commits/b2dfc6914631f76c2942895a77b699e6c98eabd3

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
jmatejcz: It was too generic and i thought it is not a good to place these tasks. i wouldn't know how to complete it myself ;p

src/rai_bench/rai_bench/tool_calling_agent/mocked_tools.py
boczekbartek: yes, I think so.

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
boczekbartek: what I mean is that the only occurence of `CheckSpawnableEntitiesTask` is here:
https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py#L510-L513

`check_spawnable_entities_val` is only used with `CheckSpawnableEntitiesTask`.
https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py#L418-L420

and `check_spawnable_entities_subtask` is only used for `check_spawnable_entities_val`.
https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py#L182-L187

I was wondering if it might be better to just move the code from [basic_tasks.py](https://github.com/RobotecAI/rai/blob/f7d36f965833d5fd5a52a91d1011d0367c30a961/src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py) to `CheckSpawnableEntitiesTask.__init__`, so that this class is not parametrized. 

I can't see how it could be parametrized with different subtask as of now, and how these subtasks could be used for other tasks, that's why I think if the code is not reusable, then it would be benefitial to keep it together.

src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py
boczekbartek: I am wondering if just a different float makes really that much difference. If we find it important don't test 10 or 100 different thresholds then? :D

src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py
boczekbartek: let's keep to as it is for now and consider in the future

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
boczekbartek: I'll add a reason:
It's not possible to use simple class property to implement abstractproperty:
```python
class Foo:
    @abstractmethod
    @property
    def a(self):
           pass
class A(Foo):
     a = "xyz" # this will cause error
```

boczekbartek (commented):
@jmatejcz 
I am not sure if it should be addressed in this PR, but while testing I've noticed an issue with how success is handled in tool calling benchmark. Please see such example:
```
robo-pc-005 ➜  rai-bench git:(jm/feat/basic-tasks-extension) ✗ tail -f src/rai_bench/rai_bench/experiments/run_2025-07-01_16-06-00/tool_calling_agent/qwen2.5:7b/0/benchmark.log
topic: /scan
type: sensor_msgs/msg/LaserScan

2025-07-01 16:06:19,328 - Benchmark logger - INFO - Running thinker
2025-07-01 16:06:19,907 - Benchmark logger - INFO - Running tool: receive_ros2_message, args: {'topic': '/pointcloud'}
2025-07-01 16:06:19,908 - Benchmark logger - INFO - Tool receive_ros2_message completed in 0.00 seconds. Tool output: {'payload': {'mock': 'payload'}, 'metadata': {'mock': 'metadata'}}
2025-07-01 16:06:19,908 - Benchmark logger - DEBUG - Tool receive_ros2_message output:

{'payload': {'mock': 'payload'}, 'metadata': {'mock': 'metadata'}}
2025-07-01 16:06:19,909 - Benchmark logger - INFO - Running thinker
2025-07-01 16:06:20,636 - Benchmark logger - INFO - Running tool: receive_ros2_message, args: {'timeout_sec': 10, 'topic': '/pointcloud'}
2025-07-01 16:06:20,637 - Benchmark logger - INFO - Tool receive_ros2_message completed in 0.00 seconds. Tool output: {'payload': {'mock': 'payload'}, 'metadata': {'mock': 'metadata'}}
2025-07-01 16:06:20,637 - Benchmark logger - DEBUG - Tool receive_ros2_message output:

{'payload': {'mock': 'payload'}, 'metadata': {'mock': 'metadata'}}
2025-07-01 16:06:20,637 - Benchmark logger - ERROR - Reached recursion limit Recursion limit of 6 reached without hitting a stop condition. You can increase the limit by setting the `recursion_limit` config key.
For troubleshooting, visit: https://python.langchain.com/docs/troubleshooting/errors/GRAPH_RECURSION_LIMIT
2025-07-01 16:06:20,638 - Benchmark logger - INFO - TASK SCORE: 1.0, TOTAL TIME: 1.737
2025-07-01 16:06:20,638 - Benchmark logger - INFO - Computing and saving average results...
2025-07-01 16:06:20,638 - Benchmark logger - INFO - Creating state based agent
2025-07-01 16:06:20,651 - Benchmark logger - INFO - State based agent created
2025-07-01 16:06:20,651 - Benchmark logger - INFO - ======================================================================================
```

Even though agent called tool correctly, it didn't stop and recursion limit was hit. In practice, this would cause `rai` agent to loop and eventually fail, potentially causing undefined behavior of the robot. 

I think it's an important one to address.

jmatejcz:
> @jmatejcz I am not sure if it should be addressed in this PR, but while testing I've noticed an issue with how success is handled in tool calling benchmark. Please see such example:
> 
> ```
> robo-pc-005 ➜  rai-bench git:(jm/feat/basic-tasks-extension) ✗ tail -f src/rai_bench/rai_bench/experiments/run_2025-07-01_16-06-00/tool_calling_agent/qwen2.5:7b/0/benchmark.log
> topic: /scan
> type: sensor_msgs/msg/LaserScan
> 
> 2025-07-01 16:06:19,328 - Benchmark logger - INFO - Running thinker
> 2025-07-01 16:06:19,907 - Benchmark logger - INFO - Running tool: receive_ros2_message, args: {'topic': '/pointcloud'}
> 2025-07-01 16:06:19,908 - Benchmark logger - INFO - Tool receive_ros2_message completed in 0.00 seconds. Tool output: {'payload': {'mock': 'payload'}, 'metadata': {'mock': 'metadata'}}
> 2025-07-01 16:06:19,908 - Benchmark logger - DEBUG - Tool receive_ros2_message output:
> 
> {'payload': {'mock': 'payload'}, 'metadata': {'mock': 'metadata'}}
> 2025-07-01 16:06:19,909 - Benchmark logger - INFO - Running thinker
> 2025-07-01 16:06:20,636 - Benchmark logger - INFO - Running tool: receive_ros2_message, args: {'timeout_sec': 10, 'topic': '/pointcloud'}
> 2025-07-01 16:06:20,637 - Benchmark logger - INFO - Tool receive_ros2_message completed in 0.00 seconds. Tool output: {'payload': {'mock': 'payload'}, 'metadata': {'mock': 'metadata'}}
> 2025-07-01 16:06:20,637 - Benchmark logger - DEBUG - Tool receive_ros2_message output:
> 
> {'payload': {'mock': 'payload'}, 'metadata': {'mock': 'metadata'}}
> 2025-07-01 16:06:20,637 - Benchmark logger - ERROR - Reached recursion limit Recursion limit of 6 reached without hitting a stop condition. You can increase the limit by setting the `recursion_limit` config key.
> For troubleshooting, visit: https://python.langchain.com/docs/troubleshooting/errors/GRAPH_RECURSION_LIMIT
> 2025-07-01 16:06:20,638 - Benchmark logger - INFO - TASK SCORE: 1.0, TOTAL TIME: 1.737
> 2025-07-01 16:06:20,638 - Benchmark logger - INFO - Computing and saving average results...
> 2025-07-01 16:06:20,638 - Benchmark logger - INFO - Creating state based agent
> 2025-07-01 16:06:20,651 - Benchmark logger - INFO - State based agent created
> 2025-07-01 16:06:20,651 - Benchmark logger - INFO - ======================================================================================
> ```
> 
> Even though agent called tool correctly, it didn't stop and recursion limit was hit. In practice, this would cause `rai` agent to loop and eventually fail, potentially causing undefined behavior of the robot.
> 
> I think it's an important one to address.

yes i think you are right, but is problem for separate issue https://github.com/RobotecAI/rai/issues/646

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
jmatejcz: Yes, i understand this, there are many tasks that can't really be validated in any other way, because it would not have much sense. I've done it this way so it is consistent across all tasks and with the subtask->validator->task architecture which might seem like an overkill sometimes. Maybe i can make `validators` param optional, and if not passed the default subtask and validator is used. What do you think?

src/rai_bench/rai_bench/tool_calling_agent/predefined/basic_tasks.py
jmatejcz: Yes, that probably true, if agent can pass this task, it doesn't make the difference what are the exact values. I think i just really wanted to make use of parameterization haha. I removed it here: https://github.com/RobotecAI/rai/pull/644/commits/f2e97574702deb72d3ddcd87bddb448e73e638d8

src/rai_bench/rai_bench/tool_calling_agent/tasks/basic.py
boczekbartek: Sounds good!

jmatejcz: i added default validators to some tasks that have very specific validators here: https://github.com/RobotecAI/rai/pull/644/commits/604f7bb9aaf61a7a28a1485274c4499e576b49be

Check it out, maybe it doesn't look the cleanest, cuz some tasks have defaults, other dont,  but the compatibility with the architecture is still there

boczekbartek (approved):

PR #651, opened 2025-07-24
fix: Transform obtained from GetROS2TransformTool can be old

### LINKED ISSUE #506: Transform obtained from GetROS2TransformTool can be old
**Describe the bug**
I've been playing with ROSBotXL demo. After I've turned off the simulation, I've queried the agent to provide its current position. The agent called GetROS2TransformTool and returned last position, not taking into account that the timestamp is from minute ago. 

**To Reproduce**
1. Run `ros2 launch examples/rosbot-xl.launch.py game_launcher:=RAIROSBotXLDemoGamePackage/RAIROSBotXLDemo.GameLauncher`
2. Run `streamlit run examples/rosbot-xl-demo.py`
3. Close 1
4. Query the agent to get it's position

**Expected behavior**
The tool should not return such old tf information. It should either mention, that the transform is old or raise an error. 

**Screenshots**

**Platform**

- OS: Ubuntu 24.04
- ROS 2 Jazzy

**Version**
877ba4afb058dc7d229d170b8f48b3fc0e9f4edc

**Additional context**

Juliaj:
This detects stale TF transform data and return warnings. 

## Purpose
This PR implements stale transform detection in the `GetROS2TransformTool` and adds proper time handling for both real-world and simulation environments.

## Proposed Changes
- Added logic in `GetROS2TransformTool` to detect stale TF transform data and include warnings in the response
- Added optional `use_sim_time` parameter to ROS2 connector initialization for proper time comparison in both real-world and simulation environments

### Additional Updates
- Updated `rosbot-xl-demo.py` to initialize connector with `use_sim_time=True`
- Added two unit tests to validate the stale transform detection functionality
- Made changes in helper.py for test cleanup. 

## Issues Addressed
- Fixes [#506](https://github.com/RobotecAI/rai/issues/506)

## Testing
- New Unit tests pass with new test coverage, no regression in existing test suite
- Validated with rosbot-xl demo execution

@maciejmajek - This is to address https://github.com/RobotecAI/rai/issues/506, your feedback on this implementation would be greatly appreciated.  To validate the changes, the original repro from the issue may not be ideal since it also stops the `/clock`.   I've added a test case to cover the simulation scenario.

### DIFF
--- a/examples/rosbot-xl-demo.py
+++ b/examples/rosbot-xl-demo.py
@@ -50,7 +50,7 @@ def initialize_agent() -> Runnable[ReActAgentState, ReActAgentState]:
         "examples/embodiments/rosbotxl_embodiment.json"
     )
 
-    connector = ROS2Connector(executor_type="multi_threaded")
+    connector = ROS2Connector(executor_type="multi_threaded", use_sim_time=True)
     tools: List[BaseTool] = [
         GetROS2TransformConfiguredTool(
             connector=connector,
--- a/src/rai_core/rai/communication/ros2/connectors/base.py
+++ b/src/rai_core/rai/communication/ros2/connectors/base.py
@@ -25,6 +25,7 @@ import rclpy.time
 from rclpy.duration import Duration
 from rclpy.executors import MultiThreadedExecutor, SingleThreadedExecutor
 from rclpy.node import Node
+from rclpy.parameter import Parameter
 from rclpy.qos import QoSProfile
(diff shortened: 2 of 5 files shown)


maciejmajek (approved):
Thank you for your contribution. Well done!

Interesting approach to shutting down the executors. This seems like a safer approach to making sure things don't go sideways. 

I appreciate the new tests too 🥇

PR #669, opened 2025-08-12
feat: nav2 blocking tool

maciejmajek:
## Purpose

Nav2 blocking tool for simpler use-cases

## Proposed Changes

`NavigateToPoseBlockingTool`

## Issues

-   Links to relevant issues

## Testing

### DIFF
--- a/src/rai_core/pyproject.toml
+++ b/src/rai_core/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "rai_core"
-version = "2.3.0"
+version = "2.4.0"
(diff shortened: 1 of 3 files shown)


Juliaj (approved):
The code change looks good to me. 
Nit, the term "blocking" might suggest the pose is blocked/obstructed, rather than the execution being synchronous. An alternative would be `NavigateToPoseSyncTool`.

maciejmajek:
Thank you, @Juliaj. I agree that the naming may be confusing.
LangChain tools support both cases: the _run (sync) and _arun (async) methods.
I wanted to reduce the naming similarity, but I’ll definitely think about it a bit more.
image

Issue #663, opened 2025-08-04
Add dataset creation from traces


rachwalk:
**Is your feature request related to a problem? Please describe.**
The data gathered through the use of `rai_benchmark` presents itself as a great opportunity to create datasets, which can be used for fine-tuning of smaller L/VLMs. Ability to create datasets automatically after a benchmark is run would be very helpful.

**Describe the solution you'd like**
A module could be created allowing for automatic creation of a dataset from a langfuse traceback after a run of `rai_benchmark` -- (however the module should not be tightly integrated with `rai_benchmark` itself -- in the future it might be beneficial to extend this functionality to non-benchmark deployments, allowing for online and/or imitation learning).

**Describe alternatives you've considered**
TBD

**Additional context**
Initial research has pointed to https://docs.unsloth.ai/ as a possible target framework to conduct fine tuning. Additional information on dataset requirements may be found there.

****Note****: Different models might require/benefit from different prompt structure for fine tuning. The dataset shouldn't necessarily take care of that, it might be better to leave it to pre-processing pipeline during fine tuning (once that's developed). However the dataset should possibly include all relevant metadata, which will allow for finetuning possibly largest range of models.

PR #673, opened 2025-08-13
feat: dataset creation from traces

Juliaj:
## Purpose

- Address https://github.com/RobotecAI/rai/issues/663 for creating dataset for model fine-tuning.

## Proposed Changes

First attempt to extract observations from Langfuse traces and format data to create a dataset for model fine-tuning. A set of model fine-tuning related helper files are created to facilitate the verification of the dataset.

[README.md ](https://github.com/RobotecAI/rai/blob/3cd3539f2b5e1e8f7640dfa65698e6de494ed69a/src/rai_finetune/README.md)contains detailed documentation for these scripts.

@rachwalk, @maciejmajek, the PR is still a work in progress and I would like to get some early feedback from you specifically on how we should format the data and whether we should align data with specific areas (e.g. tool calling).

LLM fine-tuning is new to me, your feedback for the model tuning helpers are also very much appreciated. 

## Issues

- https://github.com/RobotecAI/rai/issues/663.

## Testing

- Traces generated from `gpt-4o-mini` with Manipulation-demo have been used to verify the extraction and formatting logic.
- qwen2.5:7b-instruct has been used to test the Python files under `/model`.


src/rai_finetune/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+-   **Model Fine-tuning**: Uses Unsloth for optimized training with 4-bit quantization and LoRA support
+-   **LoRA Merger**: Merges LoRA adapter weights back into base models for standalone deployment
+-   **Ollama Converter**: Converts fine-tuned models to Ollama format using GGUF export
+
+The module is designed as a standalone package to avoid dependency conflicts between different versions of triton required by openai-whisper and unsloth-zoo.
+
+## Environment Setup
+
+This module requires Python 3.11.9 (Python 3.13 does not support Unsloth). Use pyenv to manage Python versions:
```
rachwalk: Oficially RAI supports Python 3.10 (ROS2 humble) and Python 3.12 (ROS2 jazzy). This is due to issues with wheel compatibility issues. Have you conducted tests on these releases? I think it would be beneficial for consistency reasons to have `rai_finetune` use the same python version, or at least either of the two. Of course if that is possible -- if unsloth only works on 3.11.9 then not much we can do about it.

src/rai_finetune/rai_finetune/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2025 RAI Development Team
+#
```
rachwalk:
```suggestion
# Copyright (C) 2025 Julia Jia
#
```

While I truly appreciate this, unfortunately "RAI Development Team" does not posess the legal requirements to be a copyright holder. We've had a deep read into how Apache 2.0 works, and technically the Contributor is the Copyright Holder, unless it's explicitely transferred to another Contributor. So I make this suggestion, and if you're okay being listed as the holder then please update all notices in this PR.

If these files are modified in the future we will be using the following style (example if we modify it next year):


```
# Copyright (C) 2025 Julia Jia
# Copyright (C) 2026 Robotec.AI
#
```

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
```diff
@@ -0,0 +1,489 @@
...
+import logging
+import os
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional
+
+try:
+    from langfuse import Langfuse
+except ImportError:
+    print("Error: langfuse package not found. Install with: pip install langfuse")
+    exit(1)
```
rachwalk: I am a bit confused, langfuse is listed as an obligatory dependency in pyproject.toml? Do you see it being an optional dependency in the future?

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
```diff
@@ -0,0 +1,489 @@
...
+    """Configuration for observation extraction pipeline from Langfuse"""
+
+    langfuse_host: str = "http://localhost:3000"
+    langfuse_public_key: str = ""
+    langfuse_secret_key: str = ""
+
+    # Observation related
+    output_file: str = "observations.jsonl"
+    max_observations: int = 1000
+    type_filter: str = "GENERATION"  # e.g., GENERATION, EVENT, SPAN
```
rachwalk:
```suggestion
    type_filter: Literal["GENERATION", "EVENT", "SPAN" ] = "GENERATION"  # e.g., GENERATION, EVENT, SPAN
```

If I understand correctly there is a set list of valid options.

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
```diff
@@ -0,0 +1,489 @@
...
+    start_time: Optional[str] = None  # ISO format string
+    stop_time: Optional[str] = None  # ISO format string
```
rachwalk: Perhaps it would be more user friendly to use `datetime.datetime` or `datetime.time` and convert it to ISO string when required.

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
```diff
@@ -0,0 +1,489 @@
...
+        if self.filter_models is None:
+            self.filter_models = []
+        if self.include_fields is None:
+            self.include_fields = [
+                "model",
+                "input",
+                "output",
+                "promptTokens",
+                "traceId",
+            ]
```
rachwalk: If the fields are not valid as None then please use default values, something like so. While the behaviour is the same, it helps with type hinting (e.g. Pyright won't complain about iterating over a None value).

```
from dataclasses import field
# ...
include_fields: list[str] =  field(default_factory=lambda: ["model", "input", "output", "promptTokens", "traceId"])
```

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
```diff
@@ -0,0 +1,489 @@
...
+
+        except Exception as e:
+            logger.error(f"Error fetching observations: {e}")
+            raise
+
+
+def filter_observation_by_model(item: Any, filter_models: Optional[List[str]]) -> bool:
+    """Filter observations by model value present directly or inside metadata."""
+    if not filter_models:
+        return True
```
rachwalk: Is there a use case where one would want to call this function with `filter_observation_by_model(item, None)`? I am not sure if the filter list can be elsewhere automatically generated, such that it returns `None` -- in which case I understand the None check. However, if it's unlikely I'd opt for `filter_models: List[str]` as an obligatory argument.

src/rai_finetune/README.md
rachwalk: Okay I've seen the comment in pyproject.toml and I see that 3.12 is not feasible, but 3.10 should work? Then If it's alright with you, for consistency's sake I'd keep this package officially supported for Python 3.10.

src/rai_finetune/rai_finetune/data/training_data_formatter.py
```diff
@@ -0,0 +1,937 @@
...
+                        text_content.append(
+                            f"[Image response: {item.get('image_url', {}).get('url', '')}]"
+                        )
+
+            content = " ".join(text_content)
+        elif isinstance(content, dict):
+            # Handle dictionary content (e.g., function calls)
+            content = str(content)
+
+        return {"role": "tool", "content": str(content).strip()}
```
rachwalk: I think there are some portions of these functions which could be brought to a common functionality, to avoid repetition,

src/rai_finetune/rai_finetune/data/training_data_formatter.py
```diff
@@ -0,0 +1,937 @@
...
+                    formatted_message = self.formatter.format_assistant_message(
+                        content, tool_calls, self.config
+                    )
+                elif role == "tool":
+                    formatted_message = self.formatter.format_tool_message(
+                        content, self.config
+                    )
+                else:
+                    # Skip unknown role types
+                    continue
```
rachwalk: Related to the comment above, I think a slightly more maintainable way of writing this would be to create a `self.formatted.format_message(content, self.config, role, **kwargs/tool_calls` -- though I am not dead set on it. Could you let me know why you've went with this design choice?

src/rai_finetune/rai_finetune/data/training_data_formatter.py
```diff
@@ -0,0 +1,937 @@
...
+        if current_conversation:
+            conversations.append(current_conversation)
+
+        return conversations
+
+
+class TrainingDataFormatter:
+    """Main class for formatting training data"""
+
+    def __init__(self, config: TrainingDataConfig):
```
rachwalk: I'd consider splitting this into 4 classes, each for the applicable format, instead of one class with a method for each format.

src/rai_finetune/rai_finetune/model/ollama_model_producer.py
```diff
@@ -0,0 +1,580 @@
...
+                                "properties": {},
+                                "required": [],
+                            },
+                        },
+                    },
+                ]
+            }
+
+            with open(self.output_dir / "tools.json", "w") as f:
+                json.dump(tool_definitions, f, indent=2)
```
rachwalk: Correct me if I'm wrong, but this would only be relevant for use with one particular toolset. Wouldn't it be better for the user to specify a tools.json file? If you think that'd be valid, we could add a method do automatically generate such tools.json files from RAI Agents, making it easy to setup. Let me know what you think.

rachwalk (changes requested):
Hi @Juliaj thank you for this PR. I think so far it's looking really good. I've left a few comments regarding stylistic changes. Overall I'd consider splitting some of the larger files containing multiple class definitions into separate ones -- for managability. It's a matter of preference, but I think it's better to work with smaller files, especially when large diffs come in place.

Another general comment I have would be creating abstract classes for functionalities you think could be well subclassed, i.e. Things related to specific models/providers (like langfuse may not be the only way to collect observations in the future, multiple formatters could be featured). That way we could already define an API for easy extensibility.

However to answer your question, I think that for this package and given RAI being an embodiment focused framework, I think that slight overfitting towards tool calling (and perhaps other embodying mechanisms such as the usage of `rai_whoami` in system prompts) would be beneficial.

Also thank you for your dilligent use of error reporting mechanisms and application of defensive programming, I think it's very cleanly implemented and we will be taking inspiration when it comes to development and maintance of `rai_core`.

src/rai_finetune/README.md
Juliaj: Updated by [fc2e5a814b564e](https://github.com/RobotecAI/rai/pull/673/commits/fc2e5a814b564ecc0e3f145cce4d7942b88cd86c).

src/rai_finetune/rai_finetune/__init__.py
Juliaj: Thanks for the guidance, copyright headers have been updated.

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
Juliaj: Agreed, this is redundant, removed.

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
Juliaj: Updated per your suggestion.

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
Juliaj: By friendly, did you mean to make it easier for the user to input a short value ? I have this updated to take in both ISO string and `datetime.datetime`, see https://github.com/RobotecAI/rai/blob/8f5dbfb4dc9a08df4f6260041d2329bf1e657e4f/src/rai_finetune/rai_finetune/data/extractors/base.py#L98-L99 let me know what you think.

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
Juliaj: Your point is taken. The fields are likely to be provider (Langfuse, Langsmith etc) specific, so for now I made this as an abstract method for subclass to override, see https://github.com/RobotecAI/rai/blob/8f5dbfb4dc9a08df4f6260041d2329bf1e657e4f/src/rai_finetune/rai_finetune/data/extractors/base.py#L66C1-L69C13. Wdyt?

src/rai_finetune/rai_finetune/data/langfuse_obs_extractor.py
Juliaj: updated.

src/rai_finetune/rai_finetune/data/training_data_formatter.py
Juliaj: Point taken. I have refactored this .py into a base class and a tool_calling formatter. If you could give them another look, it would be great.

src/rai_finetune/rai_finetune/data/training_data_formatter.py
Juliaj: I thought about this and introduced a new method to utllize chat template for formmating, hopefully this cuts down the complexity. https://github.com/RobotecAI/rai/blob/f16b6e81fd4b1b7e6468ec0b0346498e8ed078f2/src/rai_finetune/rai_finetune/data/formatters/base.py#L131-L147

src/rai_finetune/rai_finetune/data/training_data_formatter.py
Juliaj: See comments above, refactored the code.

src/rai_finetune/rai_finetune/model/ollama_model_producer.py
Juliaj: Very valid points. I'm planning to refactor the code and to reduce the scope of this PR, I have the fine tune helper files removed. Will refactor and include them into subsequent PR.

src/rai_finetune/README.md
```diff
@@ -0,0 +1,221 @@
+# RAI Fine-tuning Module
+
+## Module Overview
```
rachwalk: Please add in the header that this module is experimental and in development, and some features are in progress/might change working.

src/rai_finetune/pyproject.toml
```diff
@@ -0,0 +1,62 @@
+[build-system]
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry]
+name = "rai_finetune"
+version = "0.1.0"
```
rachwalk: I think it'd be better should be "0.1.0-alpha", as this is essentially a "prerelease" -- full fine tuning is not yet enabled.

rachwalk (approved):
I am approving the PR, however I left 2 comments regarding maintenance and documentation communiques. I'll be happy to merge after the comments are addressed.

maciejmajek:
+1, happy to merge after adding experimental info

Juliaj:
> +1, happy to merge after adding experimental info

@maciejmajek, I've added an experimental info header to the README and updated the version to `0.1.0-alpha` as @rachwalk suggested. Please review when you have a chance.

Issue #678, opened 2025-08-29
Grounddino weight downloaded to wrong location on error during init


pawel-kotowski:
**Describe the bug**

When I have this bug (btw why this happens? it is resolved on system restart):

```
ERROR Could not load model: CUDA unknown error - this may be due to an incorrectly set up environment, e.g. changing env variable CUDA_VISIBLE_DEVICES after program start. Setting the available devices to be zero.
```

during grounding dino init, the code decides to redownload the weights as it treats them as corrupted. Then it downloads them to a wrong location:

```
~/.cache/rai/vision/weights 13s
❯ tree
.
├── groundingdino_swint_ogc.pth
│   └── vision
│       └── weights
│           └── groundingdino_swint_ogc.pth
```

**Platform**

-   OS: Ubuntu 22.04
-   ROS 2 Version: Humble
-   Driver Version: 575.64.03      CUDA Version: 12.9

maciejmajek:
PRs attempting to fix this exact bug.
#570 
#686

maciejmajek:
Fixed by #686

PR #686, opened 2025-09-12
fix: redownload weights only on PytorchStreamReader error

pawel-kotowski:
## Purpose

-   Improve error handling in vision agents

## Proposed Changes

- Refactored paths handling - it was a bit messy and weights were redownloaded to wrong path as described in #678 
- Error resulting from broken weights file is detected via "PytorchStreamReader" message in exception (because we always get `RuntimeError` regardless of error type) in `BaseVisionAgent` class
 - Only in this case the weights are redownloaded
- Other errors are reraised, e.g. CUDA related errors which also prevent model init, even if weights file is correct

## Issues

-   [678](https://github.com/RobotecAI/rai/issues/678)

## Testing

-   Tests with broken weights files and error related to `CUDA_VISIBLE_DEVICES`

### DIFF
--- a/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
+++ b/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
@@ -41,7 +41,7 @@ class BaseVisionAgent(BaseAgent):
             self.weights_root_path / self.WEIGHTS_DIR_PATH_PART / self.WEIGHTS_FILENAME
         )
         if not self.weights_path.exists():
-            self.download_weights()
+            self._download_weights()
         self.ros2_connector = ROS2Connector(ros2_name, executor_type="single_threaded")
 
     def _load_model_with_error_handling(self, model_class):
@@ -59,13 +59,13 @@ class BaseVisionAgent(BaseAgent):
             self.logger.error(f"Could not load model: {e}")
             if "PytorchStreamReader" in str(e):
                 self.logger.error("The weights might be corrupted. Redownloading...")
-                self.remove_weights()
-                self.download_weights()
+                self._remove_weights()
+                self._download_weights()
                 return model_class(self.weights_path)
             else:
                 raise e
 
-    def download_weights(self):
+    def _download_weights(self):
         try:
             subprocess.run(
                 [
@@ -80,7 +80,7 @@ class BaseVisionAgent(BaseAgent):
             self.logger.error("Could not download weights")
             raise Exception("Could not download weights")
 
-    def remove_weights(self):
+    def _remove_weights(self):
         os.remove(self.weights_path)
 
     def stop(self):
--- a/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
+++ b/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
@@ -14,7 +14,6 @@
 
 
 import os
-import shutil
 import subprocess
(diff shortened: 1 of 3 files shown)


maciejmajek:
> Refactored paths handling - it was a bit messy and weights were redownloaded to wrong path as described in https://github.com/RobotecAI/rai/issues/678

I think this is still a problem. Interrupting on the first start causes error ".... is a directory"

pawel-kotowski:
@maciejmajek I added one minor naming fix (didn't affect anything). It works on my side, maybe @jmatejcz could check.

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
```diff
@@ -24,63 +23,65 @@
...
+            if "PytorchStreamReader" in str(e):
+                self.logger.error("The weights might be corrupted. Redownloading...")
+                self.remove_weights()
+                self.download_weights()
+                return model_class(self.weights_path)
+            else:
+                raise e
 
-    def _download_weights(self, path: Path):
+    def download_weights(self):
```
Juliaj: Is the removing of "_"  from the method naming intended ? Making this as internal/private matches the pattern of other internal method like `_load_model_with_error_handling()`

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
```diff
@@ -24,63 +23,65 @@
...
             raise Exception("Could not download weights")
 
-    def _remove_weights(self, path: str):
-        # Sometimes redownloding weights bugged and created a dir
-        # so check also for dir and remove it in both cases
-        if os.path.isdir(path):
-            shutil.rmtree(path)
-        elif os.path.isfile(path):
-            os.remove(path)
+    def remove_weights(self):
```
Juliaj: Similar to comments above, let's prefix the method name with _ to mark it as private.

Juliaj:
@pawel-kotowski, we looked at #678 during Wednesday's dev sync meeting. I was added as a reviewer since I was seeing the CUDA error somewhat frequently when running the RAI manipulation demo a few months ago. Your changes look solid overall, good work! In the PR description, you mentioned still experiencing CUDA errors even with your fixes - is that correct? I haven't been able to reproduce the error with your changes. If you're still seeing the issue, could you run the attached [cuda_error_debug.py](https://github.com/user-attachments/files/22398763/cuda_error_debug.py.txt) to help gather diagnostic information?

Juliaj:
> > Refactored paths handling - it was a bit messy and weights were redownloaded to wrong path as described in #678
> 
> I think this is still a problem. Interrupting on the first start causes error ".... is a directory"

@maciejmajek, how can we reproduce the issue you're experiencing? I tried interrupting the `RAIManipulationDemo.GameLauncher` startup with Ctrl-C several times but couldn't trigger any error.

maciejmajek (approved):
Tested again and works! Thank you!

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
maciejmajek: Done

src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
maciejmajek: Done

PR #687, opened 2025-09-15
feat: megamind context providers

jmatejcz:
## Purpose
Injecting external info to megamind prompt

## Proposed Changes

ContextProvider abstract class which has just one method get_context() which later provides an additional info during plan step

## Testing
1. import and define some context provider, like:
```python
class UsefulContext:
        def get_context(self) -> str:
            return "\nVery useful context!\n"

    useful_context = UsefulContext()
    agent = create_megamind(
        megamind_llm=supervisor_llm,
        megamind_system_prompt=task.get_system_prompt(),
        executors=executors,
        task_planning_prompt=task.get_planning_prompt(),
        context_providers=[useful_context],
    )

```
2. Run and check the prompt (easiest via Langfuse)
```bash
python3 src/rai_bench/rai_bench/examples/tool_calling_custom_agent.py 
```

### DIFF
--- a/src/rai_core/rai/agents/langchain/core/megamind.py
+++ b/src/rai_core/rai/agents/langchain/core/megamind.py
@@ -260,8 +260,8 @@ def plan_step(
 
 def create_megamind(
     megamind_llm: BaseChatModel,
-    megamind_system_prompt: str,
     executors: List[Executor],
+    megamind_system_prompt: Optional[str] = None,
     task_planning_prompt: Optional[str] = None,
     context_providers: List[ContextProvider] = [],
 ) -> CompiledStateGraph:
@@ -270,8 +270,17 @@ def create_megamind(
     Args:
         executors (List[Executor]): Subagents for megamind, each can be a specialist with
         its own tools llm and system prompt
+
+        megamind_system_prompt (Optional[str]): Prompt for megamind node. If not provided
+        it will default to informing agent of the avaialble executors and listing their tools.
+
         task_planning_prompt (Optional[str]): Prompt that helps summarize the step in a way
         that helps planning task
+
+        context_providers (List[ContextProvider]): Each ContextProvider can inject external info
+        to prompt during planning phase
+
+
     """
     executor_agents = {}
     handoff_tools = []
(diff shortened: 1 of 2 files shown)


src/rai_core/rai/agents/langchain/core/megamind.py
```diff
@@ -198,7 +207,11 @@ def get_initial_megamind_state(task: str):
     )
 
 
-def plan_step(megamind_agent: BaseChatModel, state: MegamindState) -> MegamindState:
+def plan_step(
+    megamind_agent: BaseChatModel,
+    context_providers: Optional[List[ContextProvider]],
```
boczekbartek: If the list is optional, then I think the default value of `context_providers` should be None.

Also, I prefer `List[ContextProvider] | None` syntax for typehints, but I think we don't have a consistent style in rai, so choose whichever you like, fine with the current one ;)

boczekbartek (changes requested):
@jmatejcz Good feature! Do you think it's only applicable to the `megamind` agent? I wonder if it's not more of a general feature applicable to all agents.

jmatejcz:
> @jmatejcz Good feature! Do you think it's only applicable to the `megamind` agent? I wonder if it's not more of a general feature applicable to all agents.

It is applicable for other agents for sure, but only when agent has some repeating planning node like megamind. In other cases - like react agent - this info can be just given as prompt at the beggining. I could move this class to some general destination but we don't have any other agent that could use it for now so i didn't want to create new file for 5 lines of code.
I can easily change it though if you think it should go elsewhere
@boczekbartek

boczekbartek:
@jmatejcz 
Let's leave it as it is and create an issue to investigate usefulness in detail in the future.

Could you also check [my comment](https://github.com/RobotecAI/rai/pull/687#discussion_r2348603253)? Then I think we'll be fine to merge

src/rai_core/rai/agents/langchain/core/megamind.py
jmatejcz: yea sure, applied here https://github.com/RobotecAI/rai/pull/687/commits/8dc366a9067f9a4efdd678164611ae713bf203ae

jmatejcz:
> @jmatejcz Let's leave it as it is and create an issue to investigate usefulness in detail in the future.
> 
> Could you also check [my comment](https://github.com/RobotecAI/rai/pull/687#discussion_r2348603253)? Then I think we'll be fine to merge

added issue here https://github.com/RobotecAI/rai/issues/688

boczekbartek (approved):
LGTM!

PR #692, opened 2025-09-15
ci: expect rai-core version bump on changes

maciejmajek:
## Purpose

Making sure rai-core version is always up to date.

## Proposed Changes

This CI workflow will check for changes in rai-core. If such are introduced, the CI will check if pyproject.toml has also been updated.

## Issues

-   Links to relevant issues

## Testing

CI

### DIFF
--- a/.github/workflows/require-version-bump.yml
+++ b/.github/workflows/require-version-bump.yml
@@ -37,7 +37,7 @@ jobs:
 
           # List changed files between base and head
           CHANGED_FILES=$(git diff --name-only "${BASE_SHA}" "${HEAD_SHA}")
-          echo "Changed files:\n${CHANGED_FILES}"
+          printf "Changed files:\n%s\n" "${CHANGED_FILES}"
 
           # Detect changes within src/rai_core/
           if ! printf '%s\n' "${CHANGED_FILES}" | grep -E '^src/rai_core/' >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


maciejmajek:
PR without changes to rai_core
image

PR with changes to rai_core
image

PR with changes to rai_core and pyproject.toml
image

.github/workflows/require-version-bump.yml
```diff
@@ -0,0 +1,57 @@
...
+          if ! printf '%s\n' "${CHANGED_FILES}" | grep -E '^src/rai_core/' >/dev/null 2>&1; then
+            echo "No changes detected in src/rai_core. Skipping check."
+            exit 0
+          fi
+
+          echo "Changes detected in src/rai_core. Verifying src/rai_core/pyproject.toml is updated..."
+
+          # Ensure src/rai_core/pyproject.toml is among changed files
+          if printf '%s\n' "${CHANGED_FILES}" | grep -xq 'src/rai_core/pyproject.toml'; then
+            echo "Detected change in src/rai_core/pyproject.toml."
```
boczekbartek: another grep to check if the version field has been modified would be more accurate, if package version is changed I think this grep will pass as well, right?

boczekbartek (changes requested):
Consider grepping for version

boczekbartek (approved):

PR #705, opened 2025-10-11
fix: create missing vision subdirectory for weights download path

Juliaj:
## Purpose

A regression was introduced in https://github.com/RobotecAI/rai/pull/686 which causes the following error when starting `rai_open_set_vision` agents.  

```
[python-2] FileNotFoundError: [Errno 2] No such file or directory: '/home/mmajek/.cache/rai/vision/weights/groundingdino_swint_ogc.pth'
```
The bug was missed because it only manifests when the `vision` folder doesn't exist under `.cache/rai/`.

## Proposed Changes

The `BaseVisionAgent` constructor creates `weights_root_path` however fails to create the intermediate `vision` subdirectory (WEIGHTS_DIR_PATH_PART). This causes the `wget` command to fail when attempting to download weights to a non-existent directory path.

This PR handles the directory structure creation and adds unit tests to cover the weights download. 

## Issues

-   None,

## Testing

-  New unit tests pass.
-  Remove  `rai/vision` from `.cache` and run following 
```
python src/rai_extensions/rai_open_set_vision/scripts/run_vision_agents.py
```

### DIFF
--- a/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
+++ b/src/rai_extensions/rai_open_set_vision/rai_open_set_vision/agents/base_vision_agent.py
@@ -36,10 +36,11 @@ class BaseVisionAgent(BaseAgent):
             raise ValueError("WEIGHTS_FILENAME is not set")
         super().__init__()
         self.weights_root_path = Path(weights_root_path)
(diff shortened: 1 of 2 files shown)


maciejmajek (approved):
Thank you!
Tested clean download and interruption

Issue #716, opened 2025-11-16
[CI] test_get_topics_names_and_types_tool_with_readable_topic failed on humble


Juliaj:
**Describe the bug**
The `test_get_topics_names_and_types_tool_with_readable_topic` failed because when readable=["/rosout"] was set, /rosout didn't appear in the response. The tool only showed "Writable topics:" even though /rosout was expected.

**To Reproduce**
This showed up in PR #701 CI, https://github.com/RobotecAI/rai/actions/runs/19401097295/job/55508700595?pr=701

```
 =========================== short test summary info ============================
FAILED tests/tools/ros2/test_topic_tools.py::test_get_topics_names_and_types_tool_with_readable_topic - AssertionError: assert '/rosout' in '\nWritable topics:topic: /parameter_events\ntype: rcl_interfaces/msg/ParameterEvent\n\ntopic: /tf\ntype: tf2_msgs/msg/TFMessage\n\ntopic: /tf_static\ntype: tf2_msgs/msg/TFMessage\n'
= 1 failed, 457 passed, 5 skipped, 9 xfailed, 8 warnings in 255.84s (0:04:15) ==
Error: Process completed with exit code 1.
```

**Platform**

-   OS: [e.g. Ubuntu 22.04]
-   ROS 2 Version [e.g. humble]
-   Other information you think might be relevant such as graphic card, drivers etc.

**Additional context**
It appears that the failure is intermittent (?). The current hypothesis is that on humble, /rosout may not appear in get_topic_names_and_types() unless a node is actively publishing to it. In Jazzy, it may be included by default or discovered differently.

While looking at the source code, it is interesting that readable/writable concepts are defined as a mechanism for whitelist/permission control.

maciejmajek:
This test is flaky. It fails on random occasions. I don't know the reason unfortunately. Afaik /rosout topic is always active as long the ros2 daemon is on.

PR #717, opened 2025-11-16
refactor: permission check logic for BaseROS2Tool and QoL test fixes

Juliaj:
## Purpose

- Refactor the logic to explicitly define expected behaviors for whitelisting and permission control for BaseROS2Tool
- Address https://github.com/RobotecAI/rai/issues/716.
- Reduce the test warnings.  

## Proposed Changes
- Introduce `_check_permission_and_include` in BaseROS2Tool to centralize filtering for topics/services/actions .
- Add a security model section to ROS_2_tools.md.
- QoL test fixes to reduce warnings and post shutdown exceptions

## Testing
-  New unit tests added. 
-  All unit tests were run and passed.


docs/API_documentation/langchain_integration/ROS_2_tools.md
```diff
@@ -10,6 +10,27 @@ RAI provides various ROS 2 tools, both generic (mimics ros2cli) and specific (e.
...
+    -   If `None`: all topics are readable (permissive)
+    -   If set: only topics in the list are readable (restrictive)
+    -   Note: Only applies to topics; services and actions are not checked against this parameter
+
+-   **`writable`** (Allowlist for Writing): Whitelist of topics/actions/services the agent can write/publish to
+
+    -   If `None`: all topics/actions/services are writable (permissive)
+    -   If set: only topics/actions/services in the list are writable (restrictive)
+
+-   **`forbidden`** (Blocklist): Blacklist of topics/actions/services that are always denied
```
boczekbartek:
```suggestion
-   **`forbidden`** (Blacklist): Blacklist of topics/actions/services that are always denied
```

src/rai_core/rai/tools/ros2/base.py
```diff
@@ -62,6 +62,60 @@ def is_writable(self, topic: str) -> bool:
...
+    def _check_permission_and_include(
+        self, name: str, check_readable: bool = True
+    ) -> Tuple[bool, bool, bool]:
+        """
+        Check permissions and determine if resource should be included.
+
+        Args:
+            name: Resource name (topic/service/action)
+            check_readable: If False, only checks writable (for services/actions).
+                           If True, checks both readable and writable (for topics).
```
boczekbartek: Minor formatting suggestion:
```suggestion
                            If True, checks both readable and writable (for topics).
```

docs/API_documentation/langchain_integration/ROS_2_tools.md
boczekbartek:
```suggestion
-   **`forbidden`** (Blacklist): Blacklist of topics/actions/services that are always denied
```

src/rai_core/rai/tools/ros2/generic/topics.py
```diff
@@ -164,41 +164,56 @@ def _run(self) -> str:
...
-                if self.is_writable(topic):
-                    writable_topics.append({"topic": topic, "type": type})
-
-            text_response = "\n".join(
+
+        readable_and_writable_topics: List[Dict[str, Any]] = []
+        readable_topics: List[Dict[str, Any]] = []
+        writable_topics: List[Dict[str, Any]] = []
+
+        for topic, type in topics_and_types:
```
boczekbartek: I know this part of code is not directly changed in this PR. But since it's visible here I would rename `type` -> e.g. `topic_type` as a part of this PR, not to overwrite python's builtin `type` function and avoid future problems. :)

tests/communication/ros2/helpers.py
```diff
@@ -299,35 +337,62 @@ def multi_threaded_spinner(
...
+        except Exception:
+            pass
+
+    # Give actions a moment to complete gracefully
+    time.sleep(0.15)
 
-    # Shutdown executors
+    # Cancel all timers BEFORE shutting down executors
+    # This prevents executors from trying to call timers after they're canceled
+    for node in all_nodes:
```
boczekbartek, line 363: Very cool! I think it's worth considering to contribute it to `rclpy` as well. I think this should be a part of the `executor.shutdown()` method...

boczekbartek (commented):
@Juliaj Thank you so much for cleaning up the permissions behaviour and improving tests! Looks very good.
I added a couple of minor comments.

One thing that came up in testing locally on my machine was a service call testcase fail. It only happened in the first tests run, I ran tests again 10 times and all tests passed. Do you have an idea why such behaviour might happen? 

The only connection with this PR is that this tests uses the newly implemented `_safe_spin` function [here](https://github.com/RobotecAI/rai/blob/eef8f9675a7e22cadc1559bcd29f65aaa753fa59/tests/communication/ros2/test_connectors.py#L317), so I think the issue might not be connected directly with this PR changes... 

```
============================================================================= FAILURES =============================================================================
________________________________________________________________ test_ros2_connector_create_service ________________________________________________________________
                                                                                                                                                                    
ros_setup = None, request = <FixtureRequest for <Function test_ros2_connector_create_service>>                                                                      
                                                                                                                                                                    
    def test_ros2_connector_create_service(ros_setup: None, request: pytest.FixtureRequest):                                                                        
        connector = ROS2Connector()                                                                                                                                 
        mock_callback = MagicMock()                                                                                                                                 
        mock_callback.return_value = SetBool.Response()                                                                                                             
                                                                                                                                                                    
        try:                                                                                                                                                        
            handle = connector.create_service(                                                                                                                      
                service_name="set_bool",                                                                                                                            
                service_type="std_srvs/srv/SetBool",                                                                                                                
                on_request=mock_callback,                                                                                                                           
            )                                                                                                                                                       
            assert handle is not None                                                                                                                               
            service_client = TestServiceClient()                                                                                                                    
            executors, threads = multi_threaded_spinner([service_client])                                                                                           
            service_client.send_request()                                                                                                                           
            time.sleep(0.2)                                                                                                                                         
>           assert mock_callback.called                                                                                                                             
E           AssertionError: assert False                                                                                                                            
E            +  where False = <MagicMock id='131234688348240'>.called                                                                                               
                                                                                                                                                                    
tests/communication/ros2/test_connectors.py:320: AssertionError 
```

docs/API_documentation/langchain_integration/ROS_2_tools.md
maciejmajek: Please use Blocklist or Denylist instead of Blacklist.

tests/communication/ros2/helpers.py
```diff
@@ -279,6 +301,22 @@ def publish_clock(self) -> None:
         self.publisher.publish(msg)
 
 
+def _safe_spin(executor: MultiThreadedExecutor) -> None:
```
maciejmajek: While I like that it will clean up our pytest logs, it might give a false sense that everything is working correctly. Please reconsider these changes.

tests/communication/ros2/helpers.py
maciejmajek, line 1: In the previous PR, I've removed time.sleep from the `multi_threaded_spinner` to speed up the tests. Are the time.sleep needed?

docs/API_documentation/langchain_integration/ROS_2_tools.md
Juliaj: Used Denylist.

tests/communication/ros2/helpers.py
Juliaj: I see, we don't feel particularly safe to spin this way. I had it removed.

Juliaj:
@boczekbartek, regards following

> One thing that came up in testing locally on my machine was a service call testcase fail. It only happened in the first tests run, I ran tests again 10 times and all tests passed. Do you have an idea why such behaviour might happen?

...    
def test_ros2_connector_create_service(ros_setup: None, request: pytest.FixtureRequest):                                                                                
> >           assert mock_callback.called                                                                                                                             
> E           AssertionError: assert False                                                                                                                            
> E            +  where False = <MagicMock id='131234688348240'>.called                                                                                               

I haven't seen this in my runs. This could be due to a race condition. When connector.create_service() is called, the service is created on the connector's node, but the connector's executor (running in a background thread) needs time to register it in the ROS2 graph. I added a very small delay to the test.

tests/communication/ros2/helpers.py
Juliaj: Yes, the delay alleviates the two shutdown race conditions in executor threads. I replaced one with event and reduced the sleep time for the other.

To summarize,  with the removal of `_safe_spin`, we'll see these two exceptions again intermittently. It is because the executor thread continues processing items from its wait list after they've been canceled or destroyed during shutdown.
- Timer canceled exception: The executor calls a timer that was already canceled during shutdown. The timer was canceled, but the executor hadn't removed it from its wait list yet.
- Guard condition destroyed exception: The executor triggers a guard condition that was already destroyed. This matches the earlier issue: the connector's shutdown destroys guard conditions while the test executor thread is still using them.

maciejmajek: Thank you for the deep dive. Understood!

maciejmajek (approved):

boczekbartek (approved):

PR #721, opened 2025-11-22
fix: unify use_cuda handling in GDBoxer and GDSegmenter

### LINKED ISSUE #718: rai-perception only supports partial CPU fallback
**Is your feature request related to a problem? Please describe.**
Current rai-perception only supports partial CPU fallback:

- GDSegmenter — Has automatic CPU fallback:
```
   if use_cuda:       self.device = "cuda" if torch.cuda.is_available() else "cpu"
```
Falls back to CPU if CUDA isn't available.
- GDBoxer — No automatic fallback:
```
      if not use_cuda:
       self.model = Model(self.cfg_path, self.weight_path, device="cpu")
   else:
       self.model = Model(self.cfg_path, self.weight_path)  # Uses default (CUDA)
```
If use_cuda=True and CUDA isn't available, it may fail.

- Agents — Don't expose use_cuda; they always use the default (use_cuda=True).

**Additional context**
How important is it to support CPU fallback ?

maciejmajek:
## Purpose

Closes #718 

## Proposed Changes

use_cuda handling with a fallback

## Issues

-   Links to relevant issues

## Testing

### DIFF
--- a/src/rai_extensions/rai_perception/rai_perception/vision_markup/boxer.py
+++ b/src/rai_extensions/rai_perception/rai_perception/vision_markup/boxer.py
@@ -13,10 +13,12 @@
 # limitations under the License.
 
 
+import logging
 from os import PathLike
 from typing import Dict
 
 import cv2
+import torch
 from cv_bridge import CvBridge
 from groundingdino.util.inference import Model
 from rclpy.time import Time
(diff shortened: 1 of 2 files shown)


rachwalk (approved):
LGTM

PR #724, opened 2025-11-30
chore: test rai-gsam2 with synced upstream main

Juliaj:
## Purpose

This is to test https://github.com/RobotecAI/Grounded-SAM-2/pull/5 where we pushed updated code base to test py-pi as `rai-gsam2` version 1.0.1. 

## Proposed Changes
- The main changes were in [commit c6cd8d7ac](https://github.com/RobotecAI/rai/commit/c6cd8d7ac4fdf95740cdab844d9a459dd7cb76e8) to update poetry.lock and version to rai-gsam2.
- The other changes were from https://github.com/RobotecAI/rai/pull/722. 

## Testing
- Unit tests ran and passed.
- manipulation demo was run and worked with simple prompts such as swap carrot with cube etc.
- rosbot-XL demo was run and worked with simple prompts such as "drive to kitchen and what do you see etc"

## Question: 
What else should we run to validate the upstream changes?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,7 +33,7 @@ WORKDIR /rai
 RUN git clone https://github.com/RobotecAI/rai.git .
 
 # Install Python dependencies with Poetry
-RUN poetry install --with nomad,openset
+RUN poetry install --with nomad,perception
 
 # Install ROS dependencies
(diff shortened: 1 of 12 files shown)


maciejmajek (approved):
LGTM, but please remove the test pypi source first before merging

Issue #225, opened 2024-09-13
Design SLAM RAI features


adamdbrw:
**Is your feature request related to a problem? Please describe.**
Robots often need to find out about their environment first, building a map and localizing themselves on it. During this process, RAI can be used to guide exploration for mapping or to build a semantic map during the SLAM which adds knowledge and memory, which can be used to reason about the area itself and tasks that are to be given in the area. A great start is to design solutions for RAI. 

**Describe the solution you'd like**
A document that specifies what methods can be used and how they are supposed to work. A rough effort estimation would be great, as well as proposed value estimation. This should help to prioritize work in an informed way.

**Describe alternatives you've considered**
No alternatives

**Additional context**
This could be an extensive research field.

PR #694, opened 2025-09-21
feat: improve 3d gripping point detection

Juliaj:
## Purpose

-   Address #693. 

This is joint work with @maciejmajek, who implemented the core pipeline for 3D gripping point detection.

## Proposed Changes

This PR adds a 3D gripping point detection mechanism to estimate grasping positions with various strategies for objects in their environment. 

**Pipeline Architecture:**
```
RGB + Depth Images → GroundingDINO/SAM → Point Cloud Generation → 
Filtering → Gripping Point Estimation → 3D Coordinates for gripper
```

The implementation includes:

- GetGrippingPointTool (`src/rai_core/rai/tools/ros2/detection/tools.py`): main ROS2 tool that provides gripping point detection for specified objects. This will eventually become a replacement for GetObjectPositionsTool .

- PointCloudFromSegmentation and GrippingPointEstimator (`src/rai_core/rai/tools/ros2/detection/pcl.py`) to 
   generate masked point clouds from segmented images and support multiple gripping point estimation strategies such as "centroid",  `"top_plane" and RANSAC-based plane. A set of filters are introduced to handle noisy point clouds using sklearn algorithms.

- Timeout handling (`src/rai_core/rai/tools/timeout.py`): a ThreadPoolExecutor-based timeout mechanism for long-running operations

- The detection pipeline uses a two-tier configuration design: ROS2 parameters for runtime deployment settings (topics, frames, timeouts) that can be set via launch files, and Pydantic models for algorithm-specific parameters (thresholds, strategies, clustering) that are configured at tool initialization time.

- Debugging helpers to publish to point clouds + markers for Rviz2 with an annotated image generation with projected gripping points

- Unit tests `tests/tools/ros2/test_detection_tools.py` and integration/manual tests `tests/tools/ros2/test_gripping_points.py`

- Updated examples/manipulation-demo-v2.py with the new configurable detection tool. We can remove after review.

## Issues

-   Links to #693 (3D gripping point detection feature request)

## Testing
To run unit tests
```
pytest tests/tools/ros2/test_detection_tools.py -s -v
```

To run manual tests
- Launch manipulation demo app, then run
```
pytest tests/tools/ros2/test_gripping_points.py::test_gripping_points_manipulation_demo -m "" -s -v

# or pass in strategy
pytest tests/tools/ros2/test_gripping_points.py::test_gripping_points_manipulation_demo -m "" -s -v --strategy <strategy>
```
Manipulation-demo-v2 has been run to validate the change. 

## TODOs
[...]

PR #727, opened 2025-12-03
feat: RAI Semantic Map Memory (rai_semap)

Juliaj:
## Purpose

Supports https://github.com/RobotecAI/rai/issues/225, Design SLAM RAI features

- This PR introduces `rai_semap`, a semantic map memory system that stores object detections with their 3D locations in the SLAM map frame. The system enables spatial-semantic queries like "where did I see a cup?" and provides persistent memory for robot exploration and task planning.  

## Proposed Changes
- [design doc](https://github.com/RobotecAI/rai/blob/4618982bb374a401831c28044377791b3a5dae61/src/rai_semap/design.md) & [readme](https://github.com/RobotecAI/rai/blob/4618982bb374a401831c28044377791b3a5dae61/src/rai_semap/README.md).  
- This is an early-stage implementation. The main goal is to gather feedback on the design and approach before further development.  I'm also looking for ideas to deal with following challenges, 
  - *Deduplication*: One of the main challenges is handling multiple detections of the same physical object. The system uses spatial merging with class-specific thresholds, point cloud-based matching when available, and confidence filtering to merge duplicate detections. However, distinguishing between a moved object and a new object instance remains a challenge that needs further refinement.

  - *Object Detection Accuracy*: The accuracy of object detection directly impacts the quality of the semantic map. The system includes confidence thresholds and bounding box size filtering, but detection accuracy depends on GroundingDINO.

## Testing

- New unit tests have been added.
- The current design has been validated with the rosbot-xl demo.


src/rai_semap/pyproject.toml
```diff
@@ -0,0 +1,25 @@
+[build-system]
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry]
+name = "rai_semap"
+version = "0.1.0"
+description = "Semantic mapping and spatial memory for RAI"
+authors = ["Julia Jia"]
+readme = "rai_semap/README.md"
```
maciejmajek:
```suggestion
readme = "README.md"
```

src/rai_semap/README.md
```diff
@@ -0,0 +1,133 @@
...
+For detailed design and architecture, see [design.md](../design.md).
+
+## Quick Start
+
+The following examples use the ROSBot XL demo to illustrate how to use rai_semap.
+
+### Prerequisites
+
+-   ROS2 environment set up
+-   rai_semap package installed: `poetry install --with semap`
```
maciejmajek: had to add config to the main pyproject.toml
```
diff --git a/pyproject.toml b/pyproject.toml
index d4e44d67..cba67ae0 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,13 @@ optional = true
 [tool.poetry.group.s2s.dependencies]
 rai_s2s = {path = "src/rai_s2s", develop = true, extras = ["all"]}
 
+[tool.poetry.group.semap]
+optional = true
+
+[tool.poetry.group.semap.dependencies]
+rai_semap = {path = "src/rai_semap", develop = true}
+
+
 [tool.poetry.group.simbench]
 optional = true
 ```

File: src/rai_semap/README.md [hunk 3]
```diff
@@ -0,0 +1,133 @@
...
+-   ROSBot XL demo setup, see instuctions at [rosbotxl demo](../demos/rosbot_xl.md)
```
maciejmajek:
```suggestion
-   ROSBot XL demo setup, see instuctions at [ROSBot XL demo](../../docs/demos/rosbot_xl.md)
```

src/rai_semap/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Step 0: Launch the ROSBot XL demo
+
+Follow the instruction from [rosbotxl demo](../demos/rosbot_xl.md).
```
maciejmajek:
```suggestion
Follow the instruction from [ROSBot XL demo](../../docs/demos/rosbot_xl.md).
```

src/rai_semap/README.md
```diff
@@ -0,0 +1,133 @@
...
+```python
+from geometry_msgs.msg import Point
+from rai_semap.core.backend.sqlite_backend import SQLiteBackend
+from rai_semap.core.semantic_map_memory import SemanticMapMemory
+
+# Initialize memory
+backend = SQLiteBackend("semantic_map.db")
+memory = SemanticMapMemory(
+    backend=backend,
+    location_id="rosbot_xl_demo",
```
maciejmajek: Can we stick to one location_id for all the examples?

```suggestion
    location_id="default_location",
```

src/rai_semap/design.md
```diff
@@ -0,0 +1,536 @@
...
+
+**A spatial-semantic record linking an object identity (class label, e.g., "red cup", "tool") to a 3D position (centroid/center) in the map frame, with metadata (timestamp, confidence, detection source).**
+
+Unlike pure geometric SLAM, semantic annotations enable querying "what" and "where" simultaneously. This allows agents to reason about object locations for task planning and spatial reasoning. The combination of semantic labels and 3D positions creates a different representation that bridges perception and spatial memory. The position is stored as a `Pose` object, where `pose.position` is the 3D point (computed from bounding box center or point cloud centroid when available) and `pose.orientation` is typically identity (not meaningful).
+
+```python
+# Example: Semantic annotation structure (simplified)
+{
+    "object_class": "red cup",
+    "pose": {"position": {"x": 2.5, "y": 1.3, "z": 0.8}, "orientation": {"x": 0.0, "y": 0.0, "z": 0.0, "w": 1.0}},
```
maciejmajek, line 65: Under the assumption that we run the same map with the same origin every time. I think it's ok for now, but definitely to be explored in the future to allow wider spatio-temporal-range use cases.

src/rai_semap/rai_semap/ros2/detection_publisher.py
```diff
@@ -0,0 +1,538 @@
...
+from rai_interfaces.msg import RAIDetectionArray
+from rai_interfaces.srv import RAIGroundingDino
+from rai_semap.ros2.perception_utils import enhance_detection_with_3d_pose
+
+
+# NOTE: This module contains perception layer logic that may belong to rai_perception.
+# It performs object detection and 3D pose computation, which are general perception
+# tasks not specific to semantic mapping. Consider moving to rai_perception when
+# that package has ROS2 node infrastructure in place.
+class DetectionPublisher(Node):
```
maciejmajek: Are we missing something in ROS2Connector? Was it easier to implement using Node api?

src/rai_semap/design.md
```diff
@@ -0,0 +1,536 @@
...
+
+This is the core data model that stores all semantic-spatial information in the memory system. Each annotation links a detected object to its location in the map frame, enabling spatial queries and temporal consistency tracking. The metadata field allows extensibility for point cloud features and other attributes without changing the core schema.
+
+```python
+class SemanticAnnotation:
+    id: str  # Unique identifier (UUID string)
+    object_class: str  # e.g., "red cup", "tool"
+    pose: Pose  # 3D pose in map frame (x, y, z, orientation)
+    confidence: float  # 0.0-1.0
+    timestamp: Any  # ROS timestamp of detection
```
maciejmajek: How about datetime?

maciejmajek (commented):
After pyproject.toml changes I was able to run the semap feature. Amazing work!

The design.md mentions various algorithms for spatial/temporal deduplication. Before I reflect on that, I need to read some literature on this topic. 

Since this PR is huge, I will continue the review tomorrow.

src/rai_semap/README.md
Juliaj: Thanks, checked in.

src/rai_semap/design.md
Juliaj: Very good point. Yes, current implementation supports single map per location_id with fixed origin. The use cases you suggested will support multi-session, multimap etc.

src/rai_semap/rai_semap/ros2/detection_publisher.py
Juliaj: Good point, thanks for bringing it up! I've switched `SemanticMapNode` and `DetectionPublisher` to use `ROS2Connector` . `DetectionPublisher` also uses `Node` methods directly for async service calls and raw ROS2 messages, see an [example](https://github.com/RobotecAI/rai/blob/b84524cbb0e17d8f24a5a1d851b3985bfb485dfd/src/rai_semap/rai_semap/ros2/detection_publisher.py#L419-L422). `SemanticMapVisualizer` still uses `Node` since it only publishes markers and doesn't need the connector's features, it's bit a overkill to use `ROS2Connector`.

src/rai_semap/design.md
Juliaj: I'm open to use datetime. I added some consideration [here](https://github.com/RobotecAI/rai/pull/727/commits/b84524cbb0e17d8f24a5a1d851b3985bfb485dfd), let me know your thoughts.

maciejmajek: For now, I think we can do timestamp: float - timezone naive postgresql compatible approach.

src/rai_semap/rai_semap/core/backend/spatial_db_backend.py
```diff
@@ -0,0 +1,65 @@
...
+        pass
+
+    @abstractmethod
+    def insert_annotation(self, annotation: "SemanticAnnotation") -> str:
+        """Insert annotation with spatial indexing."""
+        pass
+
+    @abstractmethod
+    def spatial_query(
+        self, center: Point, radius: float, filters: Optional[Dict[str, Any]] = None
```
maciejmajek, line 39: Generally, I feel like interwining non ROS 2 code with ROS 2 types creates confusion and limits future integrations. Consider using src/rai_core/rai/types. Such separation validates the ros2 directory you've created. Let me know what you think

src/rai_semap/rai_semap/scripts/navigate_collect.py
```diff
@@ -0,0 +1,218 @@
...
+class NavigationCollector:
+    """Navigate robot and collect detections for semantic map validation."""
+
+    def __init__(self, connector: ROS2Connector):
+        self.connector = connector
+        self.nav_toolkit = Nav2Toolkit(connector=connector)
+        self.wait_tool = WaitForSecondsTool()
+        self._nav_action_ready = False
+
+    def wait_for_nav_action_server(self, timeout_sec: float = 60.0) -> bool:
```
maciejmajek, line 39: `from rai.communication.ros2 import wait_for_ros2_actions` - I think we need to add support for timeout in ros2 waiters. The boilerplate below is what we want to eliminate with ros2connectors & their tooling

maciejmajek: But since there is no timeout in waiters, its ok as it is

src/rai_semap/rai_semap/ros2/node.py
```diff
@@ -0,0 +1,796 @@
...
+            )
+
+    def _initialize_tf(self):
+        """Initialize TF buffer and listener.
+
+        Note: ROS2Connector already creates _tf_buffer and _tf_listener internally.
+        We access the connector's TF buffer to avoid duplication.
+        """
+        # Use the connector's internal TF buffer (already created in __init__)
+        self.tf_buffer = self._tf_buffer
```
maciejmajek: Image

src/rai_semap/rai_semap/ros2/node.py
```diff
@@ -0,0 +1,796 @@
...
+                        )
+
+    def _initialize_memory(self):
+        """Initialize semantic map memory backend."""
+        database_path = self._get_string_parameter("database_path")
+        location_id = self._get_string_parameter("location_id")
+        map_frame_id = self._get_string_parameter("map_frame_id")
+        map_resolution = self._get_double_parameter("map_resolution")
+
+        backend = SQLiteBackend(database_path)
```
maciejmajek: should most likely be configurable

src/rai_semap/rai_semap/tool/query_semantic_map_tool.py
maciejmajek, line 1: Consider adding a tool that can retrieve object types previously seen in a location. 

Example: 
- The model is asked to bring an object that can appear in different shapes such as a bottle, can, or cup. 
- Finding the correct item might take three turns when using QuerySemanticMapTool, but only two turns if it can refer to a SeenObjects record. 

At some point this may require context compression, but that can be handled by a future general module.

src/rai_semap/rai_semap/ros2/node.py
```diff
@@ -0,0 +1,796 @@
...
+                    pose=pose_in_map_frame,
+                    confidence=max(existing_ann.confidence, confidence),
+                    timestamp=timestamp,
+                    detection_source=detection_source,
+                    source_frame=source_frame,
+                    location_id=self.memory.location_id,
+                    vision_detection_id=vision_detection_id,
+                    metadata=metadata if metadata else existing_ann.metadata,
+                )
+                self.memory.backend.update_annotation(updated)
```
maciejmajek: From architectural (LoD) standpoint, I think that the node should not reach that far (memory.backend). Would it make sense to expose update annotation or similar in memory?

src/rai_semap/rai_semap/core/semantic_map_memory.py
```diff
@@ -0,0 +1,402 @@
...
+        "orientation": {
+            "x": pose.orientation.x,
+            "y": pose.orientation.y,
+            "z": pose.orientation.z,
+            "w": pose.orientation.w,
+        },
+    }
+
+
+class SemanticAnnotation:
```
maciejmajek: BaseModel?

src/rai_semap/rai_semap/core/semantic_map_memory.py
```diff
@@ -0,0 +1,402 @@
...
+            "timestamp": str(self.timestamp) if self.timestamp is not None else None,
+            "detection_source": self.detection_source,
+            "source_frame": self.source_frame,
+            "location_id": self.location_id,
+            "vision_detection_id": self.vision_detection_id,
+            "metadata": self.metadata,
+        }
+
+
+class MapMetadata:
```
maciejmajek: BaseModel?

src/rai_semap/rai_semap/core/semantic_map_memory.py
```diff
@@ -0,0 +1,402 @@
...
+import uuid
+from typing import Any, Dict, List, Optional, Tuple
+
+from geometry_msgs.msg import Point, Pose
+
+from rai_semap.core.backend.spatial_db_backend import SpatialDBBackend
+from rai_semap.core.base_memory import BaseMemory
+
+
+def _pose_to_dict(pose: Pose) -> Dict[str, Any]:
```
maciejmajek: I've mentioned it in other comment, I think the Pose type from rai_core types should help us reduce boilerplate and achieve proper separation

src/rai_semap/rai_semap/ros2/node.py
```diff
@@ -0,0 +1,796 @@
...
+            )
+        elif len(msg.detections) > 0:
+            self.node.get_logger().warning(
+                f"Received {len(msg.detections)} detections but none were stored "
+                f"(confidence threshold: {confidence_threshold})"
+            )
+
+    def _process_detection(
+        self,
+        detection,
```
maciejmajek: important type missing

src/rai_semap/rai_semap/ros2/detection_publisher.py
```diff
@@ -0,0 +1,565 @@
...
+from rai_interfaces.msg import RAIDetectionArray
+from rai_interfaces.srv import RAIGroundingDino
+from rai_semap.ros2.perception_utils import enhance_detection_with_3d_pose
+
+
+# NOTE: This module contains perception layer logic that may belong to rai_perception.
+# It performs object detection and 3D pose computation, which are general perception
+# tasks not specific to semantic mapping. Consider moving to rai_perception when
+# that package has ROS2 node infrastructure in place.
+class DetectionPublisher(ROS2Connector):
```
maciejmajek: We usually use ROS2Connector through dependency injection. I'm not opposed to inheritance, just would love to know your reasoning.

src/rai_semap/rai_semap/ros2/detection_publisher.py
```diff
@@ -0,0 +1,565 @@
...
+            self.node.get_logger().debug(
+                f"Throttling: {time_since_last:.2f}s since last detection (interval: {detection_interval}s)"
+            )
+            return
+
+        self.last_image = msg
+        self.node.get_logger().debug("Processing camera image...")
+        self._process_image(msg)
+
+    def _process_image(self, image_msg: Image):
```
maciejmajek, line 392: Perhaps src/rai_extensions/rai_perception/rai_perception/tools could reduce some of the code?

maciejmajek:
@Juliaj I've finished the review. Huge feature!

src/rai_semap/rai_semap/core/backend/spatial_db_backend.py
Juliaj: > Such separation validates the ros2 directory you've created.

Indeed, time to walk the talk.

Juliaj: Nice adapter pattern: rai/types is an adapter between ROS 2 and the Python system.

src/rai_semap/rai_semap/scripts/navigate_collect.py
Juliaj: Tracked as https://github.com/RobotecAI/rai/issues/728

src/rai_semap/rai_semap/tool/query_semantic_map_tool.py
Juliaj: Would this do ? 
```
class GetSeenObjectsTool(BaseTool):
    """Tool for retrieving object types previously seen in a location."""

    name: str = "get_seen_objects"
    description: str = (
        "Get a list of object types (e.g., 'bottle', 'cup', 'table') that have "
        "been previously seen in a location. Useful for discovering what objects "
        "exist before querying for specific instances."
    )

    args_schema: Type[GetSeenObjectsToolInput] = GetSeenObjectsToolInput

    memory: SemanticMapMemory

    def _run(self, location_id: Optional[str] = None) -> str:
        """Get list of distinct object classes seen in a location."""
        pass
```
I've added skeleton code to support this ask and actual implementation will come in subsequent PR.

src/rai_semap/rai_semap/ros2/detection_publisher.py
Juliaj: Well, they're in the ros2 folder and conceptually are ROS2 connectors. Time to walk the talk of loose coupling. I went ahead and refactored it. With dependency injection, the call chain is a bit longer than direct calling.

src/rai_semap/rai_semap/core/semantic_map_memory.py
Juliaj: Power to `BaseModel`.

src/rai_semap/rai_semap/ros2/node.py
Juliaj: Magic.

Juliaj: That was left over code, removed.

src/rai_semap/rai_semap/ros2/detection_publisher.py
Juliaj: Definitely plan to. Would you be okay if I do this in subsequent PR ? I'd like to work on this in conjunction with https://github.com/RobotecAI/rai/pull/694.

src/rai_semap/design.md
Juliaj: Implemented it as `float`. I didn't put any code for back compat. In case you receive failures when running the validation script, fdisk first, i.e. `rm semantic_map.*`.

src/rai_semap/rai_semap/core/backend/spatial_db_backend.py
Juliaj: It turned out to be a bit more work than I anticipated! I ended up using both the Pose from rai.types and the ROS2 message type. See the struggle [here](https://github.com/RobotecAI/rai/blob/ae43e115a1e60d69afb772203a8fab95e5631373/src/rai_semap/rai_semap/core/semantic_map_memory.py#L24-L27)

The issue is that ROS2 transform functions (tf2_geometry_msgs) require ROS2 message types, but we also need rai.types.Pose for the Pydantic models. Using `PoseType = Any` with `arbitrary_types_allowed=True` lets Pydantic accept both, which works but isn't ideal. If you have suggestions for a cleaner approach, I'm all ears!

Juliaj:
> I've finished the review. 
@maciejmajek 

Huge thanks again for all the detailed review feedback! I've deferred a few items for follow-up—one is tracked as an issue, and the other involves moving code to rai_perception. I'll work on these in a subsequent PR to keep this one more focused.

src/rai_semap/rai_semap/tool/query_semantic_map_tool.py
maciejmajek: Yes, cool

maciejmajek (approved):
LGTM

PR #732, opened 2025-12-16
feat: initial obs implementation

maciejmajek:
## Purpose

#733 

## Proposed Changes

#733 

## Issues

#733 

## Testing

```bash
source setup_shell.sh
RAI_OBS_ENDPOINT="ws" python test_obs.py
```

### DIFF
--- a/src/rai_core/rai/observability/meta.py
+++ b/src/rai_core/rai/observability/meta.py
@@ -1,0 +1,121 @@
+# Copyright (C) 2025 Robotec.AI
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import functools
+import time
+from typing import Any, Callable, Dict
+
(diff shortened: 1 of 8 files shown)


src/rai_core/rai/observability/meta.py
```diff
@@ -0,0 +1,121 @@
...
+        schema_version = getattr(
+            cls, "__observability_schema_version__", EVENT_SCHEMA_VERSION
+        )
+        handler_map = HANDLERS.get(schema_version, {})
+
+        for method_name in methods:
+            fn = getattr(cls, method_name, None)
+            handler = handler_map.get(method_name)
+            if not fn or not handler:
+                continue
```
Juliaj, line 114: When running unit test, this caused one of the tests to fail with timeout. This in turn revealed an issue of double-wrapping inherited methods. This can happen when a base class is created with this metaclass and a subclass is created with the same metaclass. See more details at [here](https://github.com/RobotecAI/rai/blob/30c9c81b353c3f6845b94be0164d0665f95e1b2f/src/rai_core/rai/observability/meta.py#L148-L171)

Juliaj:
Since this is a WIP, I am holding off the review.  A few reference implementations to illustrate the ideas given to your RFC: [obs context](https://github.com/RobotecAI/rai/blob/6ed58a054b088cc4f0312306679531b82f3e99a0/src/rai_core/rai/observability/correlation.py#L21-L32), life cycle + heartbeat, [here](https://github.com/RobotecAI/rai/blob/6ed58a054b088cc4f0312306679531b82f3e99a0/src/rai_core/rai/agents/base.py#L42-L54)&[here](https://github.com/RobotecAI/rai/blob/6ed58a054b088cc4f0312306679531b82f3e99a0/src/rai_core/rai/observability/correlation.py#L21-L32).

PR #734, opened 2025-12-17
feat: timeout in ros2 waiters

### LINKED ISSUE #728: feat: add support for timeout in ros2 waiters
`from rai.communication.ros2 import wait_for_ros2_actions` - I think we need to add support for timeout in ros2 waiters. The boilerplate below is what we want to eliminate with ros2connectors & their tooling

_Originally posted by @maciejmajek in https://github.com/RobotecAI/rai/pull/727#discussion_r2602600763_

maciejmajek:
## Purpose

Enables the use of ros2_waiters in scenarios where the absence of certain topics, services, or actions should prevent the application from starting.

## Issues

Closes #728 

## Testing

New tests

### DIFF
--- a/src/rai_core/pyproject.toml
+++ b/src/rai_core/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "rai_core"
-version = "2.5.10"
+version = "2.6.0"
 description = "Core functionality for RAI framework"
 authors = ["Maciej Majek <maciej.majek@robotec.ai>", "Bartłomiej Boczek <bartlomiej.boczek@robotec.ai>", "Kajetan Rachwał <kajetan.rachwal@robotec.ai>"]
 readme = "README.md"
--- a/src/rai_core/rai/communication/ros2/waiters.py
+++ b/src/rai_core/rai/communication/ros2/waiters.py
@@ -12,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 import logging
 import time
-from typing import TYPE_CHECKING, Any, Callable, List, Tuple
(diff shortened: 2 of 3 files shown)


src/rai_core/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "rai_core"
-version = "2.5.10"
+version = "2.6.0"
```
Juliaj, line 7: Kajetan made good points on versioning. It may be useful to document the guidelines on how to update the version. This is what I got from LLM and we can definitely tweak these.

General guideline for updating
- Bump MAJOR when you break public APIs/behavior (remove/rename symbols, change defaults/outputs, change file formats/protocols).
- Bump MINOR when you add features in a compatible way (new APIs, new optional args, additive config, deprecations that still work).
- Bump PATCH when you only fix bugs/perf/stability/docs (no API/behavior changes expected by users).
Rule of thumb: start at PATCH; if you add features, use MINOR; if you break compatibility, use MAJOR.

Juliaj (approved):
Minor comment, LGTM.

src/rai_core/pyproject.toml
maciejmajek: > Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

https://semver.org/

I think we should classify these changes as minor.

maciejmajek: @Juliaj https://github.com/RobotecAI/rai/pull/738

PR #741, opened 2025-12-19
feat: allow passing ros2 types in send_message, service_call and start_action

maciejmajek:
## Purpose

Currently, sending ROS2 messages requires constructing a nested dictionary and passing the message type as a string. This approach has several drawbacks:
- Cumbersome for complex messages: Many ROS2 messages, such as PoseStamped, Path, or RobotTrajectory, contain multiple nested fields. Manually building dictionaries for these messages quickly becomes verbose and error-prone.
- No type checking or syntax help: Since dictionaries are untyped, IDEs and linters cannot assist with field names, types, or structure. Mistakes in key names or value types only appear at runtime.
- Hard to maintain: Any change in message structure requires carefully updating all nested dictionary fields. Large dictionaries make code difficult to read, modify, or extend.
- Prone to subtle bugs: Missing or misnamed fields may not be immediately obvious, leading to runtime errors or unexpected behavior in ROS2 communication.

Switching to using ROS2 message classes directly eliminates these problems, providing clear, type-checked, and maintainable code. The original idea of passing dictionaries instead of real message objects was intended to make it easier for AI tools or LLMs to generate code without needing to import message definitions. Both paths stay first-class: dictionaries remain the primary path for LLM tool calls, while message classes give developers type checking and IDE support. Together with the same change for `service_call` and `start_action`, RAI lowers the entry barrier for developers who have less familiarity with ROS 2, enabling them to work with messages, services, and actions in a straightforward way.

Reducing boilerplate:
```python
from rai.communication.ros2 import ROS2Connector, ROS2Context, ROS2Message
from geometry_msgs.msg import PoseStamped, Point, Quaternion

# before: requires building a large dictionary manually, requires passing the msg_type and use of ROS2Message
with ROS2Context():
    connector = ROS2Connector()
    connector.send_message(
        target="/pose_topic",
        message=ROS2Message(payload={
            "header": {
                "stamp": {"sec": 0, "nanosec": 0},
                "frame_id": "map"
            },
            "pose": {
                "position": {"x": 1.0, "y": 2.0, "z": 3.0},
                "orientation": {"x": 0.0, "y": 0.0, "z": 0.0, "w": 1.0}
            }
        }),
        msg_type="geometry_msgs/msg/PoseStamped"
    )

# after: directly using ROS2 message class, much simpler and easier to maintain, no need to pass the msg_type and use of ROS2Message
with ROS2Context():
    connector = ROS2Connector()
    pose_msg = PoseStamped()
    pose_msg.header.frame_id = "map"
    pose_msg.pose.position = Point(x=1.0, y=2.0, z=3.0)
    pose_msg.pose.orientation = Quaternion(x=0.0, y=0.0, z=0.0, w=1.0)
    
    connector.send_message(target="/pose_topic", message=pose_msg)

    # services and actions work the same way, msg_type is inferred from the Request/Goal class
    connector.service_call(target="/my_service", message=SetBool.Request(data=True))
    connector.start_action(target="/navigate_to_pose", action_data=NavigateToPose.Goal())
```

## Proposed Changes

- Allow passing ROS 2 message instances into `send_message`, `service_call` and `start_action`. `msg_type` becomes optional and is inferred from the instance.
- `msg_type` accepts the interface class as well as the string, e.g. `msg_type=SetBool`.
- Single validation path (`BaseROS2API.resolve_content`) shared by topics, services and actions. Passing a class instead of an instance, a `Response`/`Result` instead of a `Request`/`Goal`, or a `msg_type` that does not match the instance raises `ValueError`.
- Dictionaries are deep-copied before `set_message_fields`, which mutates nested lists in place (ros2/rosidl_runtime_py#33, fixed upstream only on rolling). Previously only the action path was protected.
- Fully backward compatible, all existing call shapes keep working.

## Issues

The `SetBool.Request` -> `SetBool` link turned out to be deterministic: the generated class lives in `std_srvs.srv._set_bool` as `SetBool_Request`, so the parent is derived from the module and name suffix. Same for action `Goal`.

## Testing

Extended test suite: parametrized message/service/action tests over dict, string type, class type and instance inputs, invalid input cases, and a regression test for the dict mutation.

## Release Notes

* **New Features**
  * Send pre-formed ROS2 messages directly without requiring explicit type specification
  * Type parameter is now optional when message structure is unambiguous

* **Bug Fixes**
  * Enhanced validation and error handling for invalid message inputs and missing type information

* **Documentation**
  * Updated ROS2 connector examples demonstrating message sending patterns and type inference

* **Chores**
  * Version bumped to 2.7.0

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/docs/API_documentation/connectors/ROS_2_Connectors.md
+++ b/docs/API_documentation/connectors/ROS_2_Connectors.md
@@ -34,12 +34,19 @@ The `ROS2Connector` is the main interface for publishing, subscribing, and calli
 
 ```python
 from rai.communication.ros2.connectors import ROS2Connector
+from std_msgs.msg import String
 
 connector = ROS2Connector()
 
-# Send a message to a topic
+# Send a raw ROS 2 message (msg_type is inferred)
 connector.send_message(
-    message=my_msg,  # ROS2Message
+    message=String(data="Hello"),
+    target="/my_topic"
+)
+
+# Send a message using a dictionary (msg_type is required)
+connector.send_message(
+    message={"data": "Hello"},
     target="/my_topic",
     msg_type="std_msgs/msg/String"
 )
--- a/docs/API_documentation/connectors/ROS_2_Connectors.md
+++ b/docs/API_documentation/connectors/ROS_2_Connectors.md
@@ -33,7 +33,7 @@ The `ROS2Connector` is the main interface for publishing, subscribing, and calli
 ### Example Usage
 
 ```python
-from rai.communication.ros2.connectors import ROS2Connector
+from rai.communication.ros2.connectors import ROS2Connector, ROS2Message
 from std_msgs.msg import String
 
 connector = ROS2Connector()
@@ -46,7 +46,7 @@ connector.send_message(
 
 # Send a message using a dictionary (msg_type is required)
 connector.send_message(
-    message={"data": "Hello"},
+    message=ROS2Message(payload={"data": "Hello"}),
     target="/my_topic",
     msg_type="std_msgs/msg/String"
 )
--- a/docs/API_documentation/connectors/ROS_2_Connectors.md
+++ b/docs/API_documentation/connectors/ROS_2_Connectors.md
@@ -35,6 +35,8 @@ The `ROS2Connector` is the main interface for publishing, subscribing, and calli
 ```python
 from rai.communication.ros2.connectors import ROS2Connector, ROS2Message
 from std_msgs.msg import String
+from std_srvs.srv import SetBool
+from nav2_msgs.action import NavigateToPose
 
 connector = ROS2Connector()
 
@@ -44,7 +46,7 @@ connector.send_message(
     target="/my_topic"
 )
 
-# Send a message using a dictionary (msg_type is required)
+# Send a message using a dictionary (msg_type is required, as a string or class)
 connector.send_message(
     message=ROS2Message(payload={"data": "Hello"}),
     target="/my_topic",
@@ -58,18 +60,32 @@ connector.register_callback(
     msg_type="std_msgs/msg/String"
 )
 
-# Call a service
+# Call a service with a request instance (msg_type is inferred)
 response = connector.service_call(
-    message=my_request_msg,
-    target="/my_service",
-    msg_type="my_package/srv/MyService"
+    message=SetBool.Request(data=True),
+    target="/my_service"
 )
 
-# Start an action
+# Call a service using a dictionary (msg_type is required)
+response = connector.service_call(
+    message=ROS2Message(payload={"data": True}),
+    target="/my_service",
+    msg_type=SetBool
+)
+
+# Start an action with a goal instance (msg_type is inferred)
 handle = connector.start_action(
-    action_data=my_goal_msg,
+    action_data=NavigateToPose.Goal(),
     target="/my_action",
-    msg_type="my_package/action/MyAction",
+    on_feedback=feedback_cb,
+    on_done=done_cb
+)
+
+# Start an action using a dictionary (msg_type is required)
+handle = connector.start_action(
+    action_data=ROS2Message(payload={}),
+    target="/my_action",
+    msg_type="nav2_msgs/action/NavigateToPose",
     on_feedback=feedback_cb,
     on_done=done_cb
 )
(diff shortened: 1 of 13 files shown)


maciejmajek:
@coderabbitai full review

Juliaj:
Thanks for adding the context. While I really like that your PR to make RAI friendly to human developers, I think it ought not to make it harder for LLM agents. 

- From my perspective, what makes APIs friendly for humans: static types give us autocomplete and catch errors early,  and using ROS2 message classes directly (instead of dicts) reduces boilerplate and makes code more readable. This PR drives toward that direction. That said, there's a learning curve for existing ROS2 developers who are used to working with standard ROS2 message classes, so we should keep that transition cost in mind.  

- On the flip side, I think what makes things easier for LLMs is a bit different. From your PR description I infer that schemas/ nested dicts are often simpler for LLM than constructing object hierarchies with imports. Correct me if I am wrong.

If these are true, there's a bit of a tension here. And what should we do ?

Looking at the RAI connector layer, I think we could implement a hybrid API (or dual support) to serve both audiences: developers + LLM agents.  If you like the idea, find a reference implementation [here](https://github.com/RobotecAI/rai/tree/jj/feat/hybrid_api), including the support to service/action. Basically, 

- The dual support API uses Python introspection to infer service/action types from class instances. Methods in `BaseROS2API` (`extract_service_class_from_request()` and `extract_action_class_from_goal()`) use `__qualname__` and `__module__` to extract base classes from nested Request/Response/Goal instances, making `msg_type` optional for typed (human-friendly) usage while requiring it for dict-based LLM support.

**Things you may not like**, look for the TODO in code above. Some of the arguments is overloaded and no type safety check. 


Now, does this hybrid API design violate any principles? Hmmm ... 

- On SRP: BaseROS2API does carry multiple responsibilities - message/service/action building from dicts and through class extraction, the new introspection methods we added (`extract_service_class_from_request`, `extract_action_class_from_goal` etc), type checking utilities like `is_ros2_message` etc, message conversion utilities, topic type discovery, and QoS adaptation. It looks like it violates SRP, but these are all related ROS2 utility operations. Splitting them up would create unnecessary fragmentation, so I think it's reasonable to keep them together.

- On detection logic: At first glance it looks like we're duplicating detection logic in both mixins and API classes, but they're actually complementary. Mixins use `isinstance(message, ROS2Message)` to detect wrapper objects with dict payloads - this handles connector-level type conversion between ROS2Message (wrapper) and IROS2Message (direct ROS2 classes). API classes use `_is_nested_instance()` with `__qualname__` introspection to detect nested Request/Response/Goal class instances - this handles ROS2-specific introspection to extract base service/action classes. They serve different purposes at different layers, so we're okay, I think.

- On coupling: The dependency flow is unidirectional and existed before dual support: Tools → Connectors (Mixins) → API Classes → BaseROS2API. Mixins depend on API classes for introspection methods like `extract_service_class_from_request` and `extract_action_class_from_goal`. API classes inherit from BaseROS2API for shared utilities. Tools depend on connectors for the high-level interface. No circular dependencies or inappropriate coupling here.

There are some open source projects like Pydantic which accepts both dicts and model instances in many methods, and LangChain which uses mixins for tool and agent composition. What makes RAI unique is that it combines a few patterns together that aren't often seen in other projects. We're using Protocol for type hints while using introspection to extract parent classes from nested types. We support both dicts and classes at API, connector, and tool layers. And we're working with generated classes that can't be modified, requiring structural typing.

Juliaj:
After the [additional thinking](https://github.com/RobotecAI/rai/blob/jj/feat/3dpipe_and_usability/src/rai_extensions/rai_perception/rethinking_usability.md) done in PR#750, I need to clarify my earlier statement: this change does not directly affect LLM agents, since they call publish() through the tool abstraction layer (PublishROS2MessageTool), not directly.

This change primarily benefits Extension Developers (mid-level users) who extend tools ... within the existing framework layers. It is a human-facing improvement that makes the API more usable for developers working at the component level, rather than an LLM agent-facing change. Thus, the hybrid approach is an overkill. 

> implementation for services and actions is hard.

Once I have the chance, will clean up my prototyping code for these and submit a PR.

Juliaj (approved):
Apart from rebasing, the changes look good.

PR #744, opened 2025-12-29
fix: rai-interfaces-in-dockerfile

jmatejcz:
## Purpose
rai interfaces were missing in dockerfile which resulted in some ros packages being not installed

## Proposed Changes

What does this PR add, remove or fix?

## Issues

-   Links to relevant issues

## Testing
Follow the docker setup in README. All tests should pass inside container

PR #745, opened 2025-12-30
fix: manipulation demo

jmatejcz:
## Purpose
- Fix issues with running manipulation demo on current main

## Proposed Changes
- Added missing import to deprecated conversational_agent 
- using new create_react_runnable in manipulation demo instead of deprecated conversational_agent
- fixing defining model in grounding dino

## Issues

-   Links to relevant issues

## Testing
Run manipulation demo

### DIFF
--- a/examples/manipulation-demo.py
+++ b/examples/manipulation-demo.py
@@ -16,8 +16,6 @@
 import logging
 from typing import List
 
-import rclpy
-import rclpy.qos
 from langchain_core.messages import BaseMessage, HumanMessage
 from langchain_core.tools import BaseTool
 from rai import get_llm_model
@@ -38,8 +36,6 @@ logger = logging.getLogger(__name__)
 
 
 def create_agent():
-    if not rclpy.ok():
-        rclpy.init()
     connector = ROS2Connector(executor_type="single_threaded")
 
     required_services = ["/grounded_sam_segment", "/grounding_dino_classify"]
--- a/examples/manipulation-demo.py
+++ b/examples/manipulation-demo.py
@@ -21,7 +21,7 @@ import rclpy.qos
 from langchain_core.messages import BaseMessage, HumanMessage
 from langchain_core.tools import BaseTool
 from rai import get_llm_model
-from rai.agents.langchain.core import create_conversational_agent
+from rai.agents.langchain.core import create_react_runnable
 from rai.communication.ros2 import wait_for_ros2_services, wait_for_ros2_topics
 from rai.communication.ros2.connectors import ROS2Connector
 from rai.tools.ros2.manipulation import (
@@ -38,7 +38,8 @@ logger = logging.getLogger(__name__)
 
 
 def create_agent():
-    rclpy.init()
+    if not rclpy.ok():
+        rclpy.init()
     connector = ROS2Connector(executor_type="single_threaded")
 
     required_services = ["/grounded_sam_segment", "/grounding_dino_classify"]
@@ -68,7 +69,8 @@ def create_agent():
     embodiment_info = EmbodimentInfo.from_file(
         "examples/embodiments/manipulation_embodiment.json"
     )
-    agent = create_conversational_agent(
+
+    agent = create_react_runnable(
         llm=llm,
         tools=tools,
         system_prompt=embodiment_info.to_langchain(),
--- a/src/rai_core/pyproject.toml
+++ b/src/rai_core/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "rai_core"
-version = "2.6.1"
+version = "2.6.2"
 description = "Core functionality for RAI framework"
 authors = ["Maciej Majek <maciej.majek@robotec.ai>", "Bartłomiej Boczek <bartlomiej.boczek@robotec.ai>", "Kajetan Rachwał <kajetan.rachwal@robotec.ai>"]
 readme = "README.md"


src/rai_extensions/rai_perception/rai_perception/vision_markup/boxer.py
```diff
@@ -80,15 +80,14 @@ def __init__(
...
             if torch.cuda.is_available():
                 self.device = "cuda"
             else:
                 self.logger.warning("CUDA is not available but requested, using CPU")
-                self.device = "cpu"
-        else:
-            device = "cuda" if torch.cuda.is_available() else "cpu"
-            self.model = Model(self.cfg_path, self.weight_path, device=device)
+
+        self.model = Model(self.cfg_path, self.weight_path, device=self.device)
```
Juliaj, line 89: No objection to the fix you made. I had a slightly different fix in [commit](https://github.com/RobotecAI/rai/pull/694/changes/fbcbe343a06739f582e5ea2e0f589582cf9061b5) which streamed line the if/else logic.  

We need to add unit test to catch this. If you like, you can get the change from the aforementioned commit.

src/rai_core/rai/agents/langchain/core/conversational_agent.py
```diff
@@ -69,6 +71,51 @@ def agent(
...
+
+
+def tools_condition(
+    state: State,
+    logger: logging.Logger,
+    messages_key: str = "messages",
+    connector: ROS2Connector = None,
+    manipulator_frame: str = "manipulator_base_link",
+) -> str:
+    logger.info("🔀 Running tools_condition to determine next step")
```
Juliaj: Emoji is fun! Considering this is a core module, let's remove it for now.

src/rai_core/rai/agents/langchain/core/conversational_agent.py
```diff
@@ -69,6 +71,51 @@ def agent(
...
+        logger.info(f"📝 State is list with {len(state)} items")
+    elif isinstance(state, dict) and (messages := state.get(messages_key, [])):
+        ai_message = messages[-1]
+        logger.info(f"📝 State is dict with {len(messages)} messages")
+    elif messages := getattr(state, messages_key, []):
+        ai_message = messages[-1]
+        logger.info(f"📝 State has messages attribute with {len(messages)} messages")
+    else:
+        logger.error(f"❌ No messages found in input state: {type(state)} - {state}")
+        raise ValueError(f"No messages found in input state to tool_edge: {state}")
```
Juliaj: `state` could potentially be large dict. If the error is read by a human, raw state may be too verbose, listing the key of state may be sufficient, or simply skip it.  If this is going to be read by a tool, we'll need to standardize the format of the log line, i.e. a json struct would be nice.

src/rai_core/rai/agents/langchain/core/conversational_agent.py
```diff
@@ -69,6 +71,51 @@ def agent(
...
+
+    if isinstance(state, list):
+        ai_message = state[-1]
+        logger.info(f"📝 State is list with {len(state)} items")
+    elif isinstance(state, dict) and (messages := state.get(messages_key, [])):
+        ai_message = messages[-1]
+        logger.info(f"📝 State is dict with {len(messages)} messages")
+    elif messages := getattr(state, messages_key, []):
+        ai_message = messages[-1]
+        logger.info(f"📝 State has messages attribute with {len(messages)} messages")
```
Juliaj: nit, can these log lines be consolidated ? Should these be made as debug ?

jmatejcz:
@Juliaj  Those changes to conversational agent were working on our manipulation experimental setup, so i have put it here  without putting much though. But now when i looked at it, they are unnecessary. I reverted them and left only changes to manipulation demo file itself

docker/Dockerfile
Juliaj: Based on the [discussion](https://discord.com/channels/1309212732578336808/1452653478995431497/1455587398636535871) from Discord, we want to use rosdep instead of apt to install this.

examples/manipulation-demo.py
```diff
@@ -38,7 +38,8 @@
 
 
 def create_agent():
-    rclpy.init()
+    if not rclpy.ok():
+        rclpy.init()
```
Juliaj: Any reason that we want to check if not rclpy.ok() ? As I understand, even the rclpy.init() is not strictly here needed since ROS2Connector will take care of the initialization. 

Based on what I've learned, we only need to call rclpy.init() manually if we want to control initialization parameters (e.g., domain_id, args) etc.

src/rai_core/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "rai_core"
-version = "2.6.0"
+version = "2.6.1"
```
Juliaj: You may need to bump this higher.

src/rai_extensions/rai_perception/rai_perception/vision_markup/boxer.py
```diff
@@ -80,15 +80,13 @@ def __init__(
             "vision_markup/boxer.py", "configs/gdino_config.py"
         )
         self.weight_path = str(weight_path)
-        if use_cuda:
-            if torch.cuda.is_available():
-                self.device = "cuda"
-            else:
-                self.logger.warning("CUDA is not available but requested, using CPU")
-                self.device = "cpu"
+        if use_cuda and torch.cuda.is_available():
```
Juliaj, line 83: These changes from 209a68a  shouldn't be included in this PR.  One possible reason is that your branch jm/fix/manipulation-demo is based on an older main that doesn't include 209a68a. The PR shows it because it's comparing your branch to the current main.

Rebase your branch onto the latest main so 209a68a is included in your branch history and the PR only shows your changes. Here are some of the steps 

```
# Fetch latest changes
git fetch origin

# Rebase your branch onto latest main
git rebase origin/main

# Force push the rebased branch (use --force-with-lease for safety)
git push --force-with-lease origin jm/fix/manipulation-demo
```

docker/Dockerfile
jmatejcz: okay I have separate PR for that https://github.com/RobotecAI/rai/pull/744

src/rai_extensions/rai_perception/rai_perception/vision_markup/boxer.py
jmatejcz: Mb, It was because  base branch `rai-interfaces-in-dockerfile` was not rebased. Please review it first

src/rai_core/pyproject.toml
jmatejcz: https://github.com/RobotecAI/rai/pull/745/changes/d8857e2b63685844953e36887053de543867f5da
bumped to 2.6.2

examples/manipulation-demo.py
jmatejcz: I think you are right, i tested it without and it works, removed here 
https://github.com/RobotecAI/rai/pull/745/changes/3c4a448fca9d2c334cd67547cc2b60d6dcb94bc0

src/rai_bringup/package.xml
Juliaj, line 10: nit, can you remove this blank line?

Juliaj (changes requested):
Thanks for the quick update, will. Please rebase

Juliaj (approved):
Minor comment on removing a blank line, overall looks good.

src/rai_bringup/package.xml
jmatejcz: its the format that pre-commit uses  so i m leaving it

PR #750, opened 2026-01-08
feat: redesign rai_perception API with tiered structure and improve 3D gripping point detection

Juliaj:
## Purpose

- Brings in the change from https://github.com/RobotecAI/rai/pull/694. This replaces https://github.com/RobotecAI/rai/pull/694.
- Merges the perception logic introduced in rai_semap into rai_perception([1](https://github.com/RobotecAI/rai/pull/727/changes#diff-34da25859105afd46b93e69382dbf4a6308bf42840966f8e73dbe889da71f5c5) and [2](https://github.com/RobotecAI/rai/pull/727/changes#diff-8c31dac39311638747cc904c4bd9250a8a2bf6067e350f842703bb4880eb21c7))
- Redesigns/Refactors the rai_perception API to improve development workflow and reduce cognitive load for developers.

## Proposed Changes

- Code refactor based on [Rethinking Usability](https://github.com/RobotecAI/rai/blob/jj/feat/3dpipe_and_usability/docs/extensions/rethinking_usability.md) and [API Usability Design Considerations](https://github.com/RobotecAI/rai/blob/jj/feat/3dpipe_and_usability/docs/api_design_considerations.md).
- There are breaking changes introduced, see section below for details. To reduce the impact on applications using the old service names, a legacy service names flag (`enable_legacy_service_names`) has been introduced and the default behavior is backward compatible (defaults to `true`). For new applications using only the new service names (`/detection`, `/segmentation`), you can disable legacy names by setting the flag to `false`:

  **Via launch file argument:**
  ```bash
  ros2 launch examples/manipulation-demo.launch.py game_launcher:=... enable_legacy_service_names:=false
  ```

  **Via environment variable:**
  ```bash
  ENABLE_LEGACY_SERVICE_NAMES=false ros2 run rai_perception run_perception_services
  ```

  **Via ROS2 parameter (in code):**
  ```python
  connector.node.declare_parameter("enable_legacy_service_names", False)
  ```

## Migration Guide

- [Migration Guide](https://github.com/RobotecAI/rai/blob/jj/feat/3dpipe_and_usability/src/rai_extensions/rai_perception/MIGRATION.md) have been introduced. 

## Issues

-   Links to relevant issues

## Testing

- [x] Unit tests added and ran.
- [x] manipulation demo, v1, v2. Prompt: swap any two cubes
- [x] manipulation-streamlit with both versions. Prompt:  "Place each apple on top of a cube", "Build a tower from cubes" and "Arrange objects in a line".
- [x] ROSBot - XL demo. Prompt: drive to kitchen
- [x] rai_semap
- [x] rai_bench. Test: manipulation_o3de.py started with manipulation_o3de.py without any failure.

To start manipulation-streamlit with both versions

```
# v2
AGENT_VERSION=v2 streamlit run examples/manipulation-demo-streamlit.py

# v1
AGENT_VERSION=v1 streamlit run examples/manipulation-demo-streamlit.py
```

## Summary by CodeRabbit

* **New Features**
  * Added service-based perception architecture for object detection and segmentation.
  * Introduced gripping point estimation tool with configurable filtering and estimation strategies (centroid, top-plane, biggest-plane).
  * Added debug visualization and intermediate pipeline stage publishing for perception workflows.
  * New perception presets for rapid configuration setup.

* **Documentation**
  * Added comprehensive API design considerations guide for framework extensions.
  * Added usability redesign documentation for perception module.

* **Bug Fixes & Improvements**
  * Enhanced ROS2 error handling and parameter utilities.
  * Improved test exclusion filtering to skip manual tests.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


examples/manipulation-demo.py
```diff
@@ -16,14 +16,92 @@
 import logging
 from typing import List
 
+import rclpy
 from langchain_core.messages import BaseMessage, HumanMessage
-from manipulation_common import create_agent
```
jmatejcz, line 20: why declaring create_agent here again?

jmatejcz:
Regarding rai_sim and rai_bench, I don't think a lot of changes will be required. Mainly naming changes, switching to new tools or changing paths. For example in `src/rai_bench/rai_bench/manipulation_o3de/predefined/configs/o3de_config.yaml`

maciejmajek:
I will start with coderabbit to speed things up.

@coderabbitai full review

src/rai_semap/rai_semap/ros2/node.py
```diff
@@ -25,6 +25,7 @@
 from geometry_msgs.msg import Point, Pose, PoseStamped
 from nav_msgs.msg import OccupancyGrid
 from rai.communication.ros2 import ROS2Connector
+from rai_perception.components.perception_utils import extract_pointcloud_from_bbox
```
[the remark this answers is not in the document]
Juliaj: Good catch, fixed made in commit [b170cd0](https://github.com/RobotecAI/rai/pull/750/commits/b170cd04d9243d1cdce6e9f91dea7d479b6397b4).

src/rai_extensions/rai_perception/rai_perception/components/topic_utils.py
```diff
@@ -0,0 +1,187 @@
...
+                config["camera_info_topic"],
+            ]
+            break
+
+    if required_topics is None:
+        raise RuntimeError(
+            "GetObjectGrippingPointsTool not found in tools list. "
+            "Cannot determine required topics."
+        )
+
```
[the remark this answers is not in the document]
Juliaj: Code updated to handle detection-only path, see commit [b170cd0](https://github.com/RobotecAI/rai/pull/750/commits/b170cd04d9243d1cdce6e9f91dea7d479b6397b4).

src/rai_extensions/rai_perception/rai_perception/components/service_utils.py
```diff
@@ -0,0 +1,136 @@
...
+        timeout_sec: Timeout for checking service availability (default: 0.1)
+
+    Returns:
+        True if service is available, False otherwise
+    """
+    try:
+        wait_for_ros2_services(connector, [service_name], timeout=timeout_sec)
+        return True
+    except (TimeoutError, ValueError):
+        return False
```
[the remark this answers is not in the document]
Juliaj: Handle timeout is equal or less than 0 as non-blocking, see [b170cd0](https://github.com/RobotecAI/rai/pull/750/commits/b170cd04d9243d1cdce6e9f91dea7d479b6397b4)

src/rai_extensions/rai_perception/rai_perception/components/gripping_points.py
```diff
@@ -0,0 +1,747 @@
...
+        if not self._check_min_points(pts, self.config.neighborhood_size + 1):
+            return pts
+
+        lof = LocalOutlierFactor(
+            n_neighbors=self.config.neighborhood_size,
+            contamination=self.config.outlier_fraction,
+        )
+        pred = cast(NDArray[np.int64], lof.fit_predict(pts))  # type: ignore[no-any-return]
+        mask = pred == 1  # 1 = inlier, -1 = outlier
+        return self._apply_mask_or_fallback(pts, mask)
```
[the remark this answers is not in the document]
Juliaj: Good catch, fix made.

src/rai_extensions/rai_perception/rai_perception/components/gripping_points.py
```diff
@@ -0,0 +1,747 @@
...
+
+
+class PointCloudFromSegmentationConfig(BaseModel):
+    box_threshold: float = Field(
+        default=0.35, description="Box threshold for GDINO object detection"
+    )
+    text_threshold: float = Field(
+        default=0.45, description="Text threshold for GDINO object detection"
+    )
+
```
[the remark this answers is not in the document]
Juliaj: Timeout added.

src/rai_extensions/rai_perception/rai_perception/components/detection_publisher.py
```diff
@@ -391,11 +400,15 @@ def image_callback(self, msg: Image):
...
-                "DINO service not ready, skipping detection"
+        dino_service = self._get_string_parameter("dino_service")
+        # Ensure service is available before use
+        if not check_service_available(self.connector, dino_service, timeout_sec=0.1):
+            raise ROS2ServiceError(
+                service_name=dino_service,
+                timeout_sec=0.1,
+                service_state="unavailable",
+                suggestion="DINO service not ready. Check if detection service is running.",
             )
```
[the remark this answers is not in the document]
Juliaj: Code updated, see https://github.com/RobotecAI/rai/pull/750/commits/b170cd04d9243d1cdce6e9f91dea7d479b6397b4.

examples/manipulation-demo.py
Juliaj: The reason is that the agent is going to invoke the new tool. I've consolidated the code into examples/manipulation_common.py.

Juliaj:
> Regarding rai_sim and rai_bench, I don't think a lot of changes will be required. Mainly naming changes, switching to new tools or changing paths. For example in `src/rai_bench/rai_bench/manipulation_o3de/predefined/configs/o3de_config.yaml`

@jmatejcz thanks for help look into this. The current code changes are back-compat by default. Once we have this merged, I will look into migrate existing code to the new service names.

src/rai_core/rai/tools/timeout.py
```diff
@@ -0,0 +1,161 @@
...
+F = TypeVar("F", bound=Callable[..., Any])
+
+
+class RaiTimeoutError(Exception):
+    """Custom timeout exception for RAI tools"""
+
+    pass
+
+
+def timeout(seconds: float, timeout_message: str | None = None) -> Callable[[F], F]:
```
jmatejcz, line 49: did you test/use it anywhere?

src/rai_extensions/rai_perception/rai_perception/components/exceptions.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import Any, Optional
+
+
+class PerceptionError(Exception):
+    """Base exception for all perception-related errors."""
+
+    pass
+
+
+class PerceptionAlgorithmError(PerceptionError):
```
jmatejcz: These errors are not used anywhere

jmatejcz (commented):
Look into to code, just left some minor questions.

How can i test this whole new functionality? manipulation demo v2 is enough? And what to look for ?

src/rai_core/rai/tools/timeout.py
Juliaj: It is used in code. In src/rai_extensions/rai_perception/rai_perception/tools/gripping_points_tools.py (line 184): used as a decorator on _run_with_timeout() inside the _run method of GetObjectGrippingPointsTool.
```
    def _run(self, object_name: str, debug: bool = False) -> str:
        @timeout(
            self.timeout_sec,
            f"Gripping point detection for object '{object_name}' exceeded {self.timeout_sec} seconds",
        )
        def _run_with_timeout():
```

There is a unit test that covers the timeout decorator usage: test_get_gripping_point_tool_timeout in tests/rai_perception/tools/test_pcl_detection_tools.py.

src/rai_extensions/rai_perception/rai_perception/components/exceptions.py
Juliaj: Good catch, only the `PerceptionError` is used, I'll have all of them removed.

Juliaj: Code updated, see https://github.com/RobotecAI/rai/pull/750/commits/03eb766cc3cff2759427ea07efcbaf36b0b91c40

Juliaj:
> How can i test this whole new functionality? manipulation demo v2 is enough? And what to look for ?

@jmatejcz, thanks for looking into this. The testing I have done is listed in the PR description. A few additional things perhaps
- Rerun the manipulation demo v2 and rai_bench sanity check test to make sure I didn't miss there.
- Some work of this PR is for usability, you could evaluate it with a potential feature for the overall developer experience. For example, does the new structure make it easier to find the folder to make the change? I did some initial evaluation with [this new use case](https://github.com/RobotecAI/rai/blob/jj/feat/3dpipe_and_usability/docs/extensions/rethinking_usability.md#new-use-case-switching-between-detection-models).
- A check on another internal demo would be nice as well if anyone has the bandwidth for backward compatibility.

jmatejcz:
@Juliaj  I tested V1 and V2 manipulaiton demo with same prompts. They both work and results are the same.
Tested Rosbot demo, it also works

src/rai_extensions/rai_perception/rai_perception/tools/gdino_tools.py
```diff
@@ -90,22 +92,78 @@ def _run(self, *args, **kwargs):
         """Abstract method - must be implemented by subclasses."""
         raise NotImplementedError("Subclasses must implement _run method")
 
+    def _get_detection_service_name(self) -> str:
```
jmatejcz, line 105: I started my dive into this refactor with the tools/ dir as it is highest tier layer in this API. 

If I understand correctly from the documents you wrote and the code, now switching detection models is possible by setting ros param.

In such case i would rename the base tool and its varaibles, as sticking with "GroundingDino..." names suggest to user that this work only with Grounding dino model. I think renaming to `DetectionBaseTool` etc. would increase role expressiveness

src/rai_extensions/rai_perception/rai_perception/components/gripping_points.py
```diff
@@ -0,0 +1,752 @@
...
+    neighborhood_size: int = Field(
+        default=20,
+        description=(
+            "Number of neighbors to consider for local density estimation. "
+            "Maps to Local Outlier Factor n_neighbors when strategy='conservative_outlier_removal'"
+        ),
+    )
+
+
+def _publish_gripping_point_debug_data(
```
jmatejcz: it shouldn't be "private" method I believe

src/rai_extensions/rai_perception/rai_perception/tools/gdino_tools.py
jmatejcz: The same applies to other fragments in code when "dino..." naming is used , for example "dino"node"

jmatejcz (commented):
I did a deeper dive into the changes. Thank you for the documents you wrote on the design - they are great source of knowlegde and also give an idea about your approach to this refactor.

In my opition the layered approach  is a very good idea. The `tools/` dir is very easy to use as it supposed to be.
Also the `components/` are designed and self explenatory. I like the idea with presets, especially  with the comments which is best for what. 

These  changes for sure reduce friction  for new users and I love the abstraction level that they introduce. I tried myself using some of this code and I didn't have to even look at  `algorithms/` or implementations of components to use the module.

Now i can see the issue that rest of rai modules are not organized in such a way,  which might confuse user, but I guess we can refactor them in the future too ;p

src/rai_extensions/rai_perception/rai_perception/tools/gdino_tools.py
Juliaj: @jmatejcz, thanks a bunch for the feedback. This is a good catch.

In addtion to this module, a few other modules (components/gripping_points.py, tools/segmentation_tools.py) are also tightly coupled to GroundingDINO specifics, e.g. dardcoded RAIGroundingDino service type, and model-specific parameters (box_threshold, text_threshold). Overall, the refactoring change needs a careful evaluation. 

Given the current PR scope, I've added a note for this in MIGRATION.md under "Generic Detection Tools Abstraction" so that we may address in a follow-up PR after merge. Hope you are okay with deferring this.

src/rai_extensions/rai_perception/rai_perception/components/gripping_points.py
Juliaj: Thanks for the suggestion.

I'm neutral on this. Currently, developers can access this via `GetObjectGrippingPointsTool` after setting debug=True in the tool's input, which publishes the debug data automatically.

Making it public would allow standalone reuse of the visualization logic, but since it's a debug-only utility with performance overhead, keeping it private help avoid committing to a debug-only public API. What do you think ?

src/rai_extensions/rai_perception/rai_perception/tools/gdino_tools.py
jmatejcz: okay, so the main problem is that interface of grounding dino differs from other models?
And now we can't actually use any model besides GD, but if rai interfaces are added for other models, this refactor provides a fundamentals for next refactor?

jmatejcz: because now, parts of code enable model switching, but it does not actually work yet yes?

src/rai_extensions/rai_perception/rai_perception/components/gripping_points.py
jmatejcz: If the intention is to use it only in `GetObjectGrippingPointsTool` then it should be moved to `get_gripping_points_tool.py`. i think it breaks encapsulation and is a bit weird  to have a private method in file that is not even used in the same module.

But it can be made "public" if the intention is for user to use it. What do you think?

src/rai_extensions/rai_perception/rai_perception/tools/gdino_tools.py
Juliaj: Correct. This PR laid out the foundation for model switching. For example, the DetectionService has infrastructure for model switching.  In base_vision_service.py, following method reads the parameter and initialize the model, https://github.com/RobotecAI/rai/blob/65ddb05b50ab86ddc321a0b27cb39d7d001d091f/src/rai_extensions/rai_perception/rai_perception/services/base_vision_service.py#L96-L109


But the service interface is still GroundingDINO-specific, so it cannot actually switch to other models yet. In addition, tools remain tightly coupled to GroundingDINO's interface.

src/rai_extensions/rai_perception/rai_perception/components/gripping_points.py
Juliaj: Moved the method, [commit](https://github.com/RobotecAI/rai/pull/750/changes/2ce950d3f80d4d171baec5a3f7961692fc2fe931).

src/rai_extensions/rai_perception/rai_perception/tools/gdino_tools.py
jmatejcz: okay, that you for clarification. Do you think we should leave some note , that only GD is currently supported? in this method  above for example

jmatejcz: @Juliaj thank you for this PR and clarification. I approve it, you could add the note or something add we are done ;p

jmatejcz (approved):

src/rai_extensions/rai_perception/rai_perception/tools/gdino_tools.py
Juliaj: Thanks @jmatejcz! I've updated module comments across relevant files to clarify that model switching is not yet fully implemented and to document the current coupling to GroundingDINO/Grounded SAM interfaces, along with future directions.

I evaluated the effort required for true model-agnostic refactoring (generic service interfaces). It's medium to high complexity due to parameter abstraction, backward compatibility, and refactoring across services, tools, and components. 

As an alternative, we can use model-specific services and tools per model, which may be simpler and faster to implement but may result in some code duplication. I've updated MIGRATION.md with the trade-offs between these approaches.

Issue #753, opened 2026-01-21
ROSbot XL demo process given prompts. Help.


kaanoz1:
As a group we are trying to run RAI in order to accomplish our project. As for testing and initialize our project we decided to launch the demo of launch [ROSbot XL demo](https://robotecai.github.io/rai/demos/rosbot_xl/). We can completely launch the game engine and GUI and allow us control to robot by prompts without any errors. But for some reason our robot rejects EVERY prompt that we entered and it is impossible to move our robot by prompts. I will send the logs. May anybody help?

```bash
libgomp: Invalid value for environment variable OMP_WAIT_POLICY
2026-01-21 13:15:38 c91fe47d0149 matplotlib.font_manager[29258] INFO Failed to extract font properties from /usr/share/fonts/truetype/noto/NotoColorEmoji.ttf: Can not load face (unknown file format; error code 0x2)
2026-01-21 13:15:38 c91fe47d0149 matplotlib.font_manager[29258] INFO generated new fontManager

libgomp: Invalid value for environment variable OMP_WAIT_POLICY
2026-01-21 13:15:48 c91fe47d0149 rai.initialization.model_initialization[29258] INFO Initializing complex_model: Vendor: ollama, Model: llama3.2
2026-01-21 13:16:56 c91fe47d0149 ROS2Connector[29258] WARNING MultiThreadedExecutor is 66.43 seconds behind. If you see this message frequently, consider switching to SingleThreadedExecutor.
2026-01-21 13:17:21 c91fe47d0149 httpx[29258] INFO HTTP Request: POST http://10.10.100.61:11434/api/chat "HTTP/1.1 200 OK"
2026-01-21 13:17:21 c91fe47d0149 rai.agents.langchain.core.tool_runner[29258] INFO Running tool: navigate_to_pose, args: {'x': -0.2175, 'y': -0.8775, 'yaw': None, 'z': None}
2026-01-21 13:17:21 c91fe47d0149 rai.agents.langchain.core.tool_runner[29258] INFO 
                                    Validation error in tool navigate_to_pose:
                                    NavigateToPoseToolInput
                                    Number of errors: 2
                                    Errors:
                                    [
  {
    "type": "float_type",
    "loc": [
      "z"
    ],
    "msg": "Input should be a valid number",
    "input": null
  },
  {
    "type": "float_type",
    "loc": [
      "yaw"
    ],
    "msg": "Input should be a valid number",
    "input": null
  }
]
                                
[...]

PR #757, opened 2026-01-26
fix: NavigateToPose.Result has no attribute error_code and log error msg

Juliaj:
## Purpose

Fix  
- https://github.com/RobotecAI/rai/issues/753#issuecomment-3792135950
```
"navigate_to_pose_blocking", error: 'NavigateToPose_Result' object has no attribute 'error_code'
```
- https://discord.com/channels/1309212732578336808/1309212733039841304/1465025297203204279
```
"navigate_to_pose_blocking", error: 'NavigateToPose_Result' object has no attribute 'error_message'
```
## Proposed Changes

- `send_goal()` returns `GetResultService.Response` instead of `NavigateToPose.Result`.  It is not clear on why during runtime, it would report the attribute error of `error_code`. Added debug logging to investigate further.
- To get additional error info, checking status and then error message may be more descriptive. The nav2 has limited support for error_code. `_get_error_message` is to support this. 
- In addition to the new tests, the changes in tests are to address https://github.com/RobotecAI/rai/issues/759 and time to msg handling for humble.

## Issues

-   Links to relevant issues

## Testing

-   Unit test added.

### DIFF
--- a/src/rai_core/pyproject.toml
+++ b/src/rai_core/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "rai_core"
-version = "2.8.1"
+version = "2.8.2"
 description = "Core functionality for RAI framework"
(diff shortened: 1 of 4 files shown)


rachwalk (approved):
LGTM

PR #766, opened 2026-03-03
perf: parallelize perception service initialization

KE7:
## Purpose

Reduce startup time for perception services and minimize log spam during initialization.

## Proposed Changes

- Uses threading to load GroundingDINO and Grounded SAM models concurrently, significantly reducing startup time.
- Refines readiness logging in `o3de_bridge.py` to only log missing components every 5 seconds (instead of 0.5s) to avoid log spam, while adding a note about initial weight download delays.
- Removes hardcoded weight size estimation from warning messages.

## Issues

- None

## Testing

- Manually verified perception services start correctly and concurrently
- Verified log output is cleaner during initialization


maciejmajek:
@KE7 
Thanks! LGTM
Please run `pre-commit run -a`, I will merge it

maciejmajek (approved):

PR #772, opened 2026-03-18
feat: http connector

rachwalk:
## Purpose

RAI could benefit from ability to utilize the HTT protocol

## Proposed Changes

Adds a connector for creating and calling HTTP services

## Issues

N/A

## Testing

tests were added for all relevant functionality

### DIFF
--- a/src/rai_core/pyproject.toml
+++ b/src/rai_core/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "rai_core"
-version = "2.11.0"
+version = "2.11.0.a1"
 description = "Core functionality for RAI framework"
 authors = ["Maciej Majek <maciej.majek@robotec.ai>", "Bartłomiej Boczek <bartlomiej.boczek@robotec.ai>", "Kajetan Rachwał <kajetan.rachwal@robotec.ai>"]
 readme = "README.md"
@@ -29,6 +29,7 @@ langchain-ollama = "^0.3.4"
 langchain-google-genai = "*"
 langchain-community = "*"
 
+aiohttp = "^3.13.3"
 requests = "^2.32.2"
 coloredlogs = "^15.0.1"
 tqdm = "^4.66.4"
--- a/src/rai_core/pyproject.toml
+++ b/src/rai_core/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "rai_core"
-version = "2.10.0"
+version = "2.11.0"
 description = "Core functionality for RAI framework"
 authors = ["Maciej Majek <maciej.majek@robotec.ai>", "Bartłomiej Boczek <bartlomiej.boczek@robotec.ai>", "Kajetan Rachwał <kajetan.rachwal@robotec.ai>"]
 readme = "README.md"
--- a/src/rai_core/rai/communication/http/api.py
+++ b/src/rai_core/rai/communication/http/api.py
@@ -13,11 +13,10 @@
 # limitations under the License.
 
 import asyncio
+import logging
 import threading
 from enum import IntFlag
-from typing import Callable, Optional, Any
-import json
-import logging
+from typing import Callable, Optional
 
 import aiohttp
 from aiohttp import ClientSession, ClientTimeout, web
@@ -153,4 +152,3 @@ class HTTPAPI:
             except Exception as e:
                 logging.warning(f"Background request failed or timed out: {e}")
         self.unresolved_futures.clear()
-
--- a/src/rai_core/rai/communication/http/api.py
+++ b/src/rai_core/rai/communication/http/api.py
@@ -1,0 +1,156 @@
+# Copyright (C) 2026 Kajetan Rachwał
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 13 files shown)


src/rai_core/rai/communication/http/api.py
```diff
@@ -0,0 +1,154 @@
...
+        self.app = web.Application()
+        self.runner = web.AppRunner(self.app)
+        self.loop = asyncio.new_event_loop()
+        self.client_session = None
+        self._thread = threading.Thread(target=self._start_loop, daemon=True)
+        self._started_event = threading.Event()
+        self.unresolved_futures = []
+
+        self.websockets: dict[str, set[web.WebSocketResponse]] = {}
+        self.ws_clients: dict[str, aiohttp.ClientWebSocketResponse] = {}
```
Juliaj, line 56: Here everything runs through a single `HTTPAPI` object and keeps setup simple -- one class, one event loop, one server. For a REST-only use case this is straightforward and easy to reason about. 

Looking at HTTPHRIConnector class, it implies for message sequencing support which generally websocket handles better. One alternative is to split this to support two different modes: REST is stateless (send request, get response, done). WebSocket is stateful (open a connection, keep it alive, push messages in both directions). 

This also helps simplify shutdown logic(REST, stop accepting requests versus closing open sockets for websocket etc) and error recovery.

src/rai_core/rai/communication/http/api.py
```diff
@@ -0,0 +1,154 @@
...
+                    and self.client_session is not None
+                ):
+                    await self.client_session.close()
+                if self.mode & HTTPConnectorMode.server:
+                    await self.runner.cleanup()
+                self.loop.stop()
+
+            asyncio.run_coroutine_threadsafe(_shutdown(), self.loop)
+
+        shutdown()
```
Juliaj, line 91: This `shutdown()` doesn't guanrantee a wait for `asyncio.run_coroutine_threadsafe(_shutdown(), self.loop)` which returns immmediately.

src/rai_core/rai/communication/http/api.py
```diff
@@ -0,0 +1,154 @@
...
+
+        self.routes: dict[str, list[str]] = {}
+
+        self.app = web.Application()
+        self.runner = web.AppRunner(self.app)
+        self.loop = asyncio.new_event_loop()
+        self.client_session = None
+        self._thread = threading.Thread(target=self._start_loop, daemon=True)
+        self._started_event = threading.Event()
+        self.unresolved_futures = []
```
Juliaj, line 53: We may want to make this as fixed capacity so that it doesn't grow unbounded. For long lived process, this may cause memory leak.

src/rai_core/rai/communication/http/api.py
```diff
@@ -0,0 +1,154 @@
...
+        headers: Optional[dict],
+        **kwargs,
+    ) -> tuple[str, int]:
+        if not (self.mode & HTTPConnectorMode.client):
+            raise HTTPAPIError("Tried sending request with client mode disabled!")
+        timeout_cfg = ClientTimeout(timeout) if timeout is not None else None
+        if payload is not None and "json" not in kwargs and "data" not in kwargs:
+            kwargs["json"] = payload  # aiohttp will set appropriate headers
+        if headers is not None:
+            kwargs["headers"] = {"Content-Type": "application/json"}
```
Juliaj, line 131: This overwrites caller's headers, instead of throwing them away, should we consider forwarding them ?

src/rai_core/rai/communication/http/messages.py
```diff
@@ -0,0 +1,29 @@
...
+
+from rai.communication.base_connector import BaseMessage
+from rai.communication.hri_connector import HRIMessage
+
+
+class HTTPMessage(BaseMessage):
+    headers: Optional[dict] = None
+    method: Literal[
+        "GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE", "POST", "PATCH", "CONNECT"
+    ]
```
Juliaj, line 25: TRACE has security implications. We can take this out and add it back once there is use case.

src/rai_core/rai/communication/http/connectors/base.py
```diff
@@ -0,0 +1,134 @@
...
+from aiohttp import web
+
+from rai.communication import BaseConnector
+from rai.communication.http.api import HTTPAPI, HTTPConnectorMode
+from rai.communication.http.messages import HTTPMessage
+
+T = TypeVar("T", bound=HTTPMessage)
+
+
+class HTTPBaseConnector(BaseConnector[T]):
```
Juliaj, line 26: If we decide to go with websocket mode support, consider add api such as `send_message`, `register_callback` etc for streaming support.

src/rai_core/rai/communication/http/connectors/hri_connector.py
```diff
@@ -0,0 +1,77 @@
...
+            "images": [HTTPHRIMessage._image_to_base64(img) for img in message.images],
+            "audios": [HTTPHRIMessage._audio_to_base64(aud) for aud in message.audios],
+            "message_author": message.message_author,
+            "communication_id": message.communication_id,
+            "seq_no": message.seq_no,
+            "seq_end": message.seq_end,
+        }
+
+    @staticmethod
+    def payload_to_message(data: dict) -> HTTPHRIMessage:
```
Juliaj, line 64: `message_to_payload / payload_to_message` are the right abstractions but these can't be called from HTTPBaseConnector.service_call. We may need to add hooks to base.py to allow subclass customization in order to make these callable.

src/rai_core/pyproject.toml
maciejmajek, line 1: We need aiohttp dep here

maciejmajek (commented):
review wip

PR #774, opened 2026-03-20
fix: make vendor config sections optional in config.toml

### LINKED ISSUE #773: Configuration requires unused options
This function used for loading: https://github.com/RobotecAI/rai/blob/59d0e8965e1c7d455ff2d52d404415646a7399d7/src/rai_core/rai/initialization/model_initialization.py#L99 requires that all vendors have defined options, which is fine when using a default configuration, but a bit bothersome when creating a custom deployment.

To illustrate:
```
[vendor]
simple_model = "ollama"
complex_model = "ollama"
embeddings_model = "ollama"

[ollama]
simple_model = "qwen3.5:9b"
complex_model = "qwen3.5:9b"
embeddings_model = "qwen3.5:9b"
base_url = "http://localhost:11434"
```

This config fails to load, even though I am certain that I will not be using any other vendor. I think it would be beneficial to support such minimal configs for users.

Bingtagui404:
## Summary

- Make `[aws]`, `[openai]`, `[ollama]`, `[google]`, and `[tracing]` sections optional in `config.toml`
- Only the sections for vendors actually referenced in `[vendor]` are required
- Missing sections get safe placeholder defaults so single-vendor setups work

closes #773


maciejmajek (approved):
LGTM, thanks!

PR #777, opened 2026-03-24
chore(mani_bench): Simplify toolset

maciejmajek:
## Purpose

The o3de manipulation benchmark should evaluate manipulation skills through reasoning, without requiring additional tool calls to retrieve a topic list or fetch an image.

The image tool should be preconfigured so the model can focus on the task itself.

While this removes part of the benchmark (i.e., fewer steps to reason about), it better aligns the evaluation with what is actually being tested.

## Proposed Changes

Update the available tools in the o3de manipulation benchmark.

## Testing

-   Running benchmark


maciejmajek:
@KE7 fyi, can't add you as a reviewer

jmatejcz (approved):
Good change

PR #794, opened 2026-06-19
chore: remove unused sumbodule

maciejmajek:
## Purpose

  A stray submodule gitlink (`src/src/rai_interfaces`, mode `160000`) was accidentally introduced in #787.

  The entry is inert under normal development but breaks any tooling that clones the repo and runs `git submodule update --init` (e.g. pip/uv when installing a package directly from this gitbranch).

  ## Proposed Changes

  - Removed the dangling `src/src/rai_interfaces` submodule gitlink from the tree

  ## Issues

  - N/A

  ## Testing

  - `pixi install` with `rai-perception` resolved via this branch now completes successfully - `git submodule update --init` no longer fails with `fatal: No url found for submodule path
  'src/src/rai_interfaces' in .gitmodules`


rachwalk (approved):
LGTM

PR #817, opened 2026-07-21
fix(tools): reject non-positive ROS2 CLI command timeout

Bartok9:
### Why
`run_with_timeout` forwarded any `timeout_sec` into `threading.Timer`. Values `<= 0` give a meaningless or immediate kill race; `bool` subclasses `int` so `False` was quietly treated as 0.

### What
- Reject non-numeric/`bool` timeouts with `TypeError`
- Reject `timeout_sec <= 0` with `ValueError`
- Offline unit tests load `cli.py` without requiring ROS2 sourced (package `__init__` gates on ROS_DISTRO)

### Tests
- [x] `PYTHONPATH=src/rai_core .venv/bin/python -m pytest tests/tools/test_ros2_cli_timeout.py -q` (5 passed)

Shell forbidden-character safety unchanged. Human-reviewed micro-diff.

### DIFF
--- a/src/rai_core/rai/tools/ros2/cli.py
+++ b/src/rai_core/rai/tools/ros2/cli.py
@@ -22,6 +22,12 @@ FORBIDDEN_CHARACTERS = ["&", ";", "|", "&&", "||", "(", ")", "<", ">", ">>", "<<
 
 
 def run_with_timeout(cmd: List[str], timeout_sec: int):
+    if isinstance(timeout_sec, bool) or not isinstance(timeout_sec, (int, float)):
+        raise TypeError(
+            f"timeout_sec must be a positive number, got {type(timeout_sec).__name__}"
+        )
+    if timeout_sec <= 0:
+        raise ValueError(f"timeout_sec must be positive, got {timeout_sec!r}")
     proc = Popen(cmd, stdout=PIPE, stderr=PIPE)
     timer = Timer(timeout_sec, proc.kill)
     try:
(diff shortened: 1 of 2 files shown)


maciejmajek (changes requested):
Thank you for the contribution! While I like the fix, the `def _load_cli_module():` fixture is pure dark magic. If such structure is strictly required, I propose keeping this path untested.

Bartok9:
Thanks for the review @maciejmajek — totally fair, that fallback was doing too much. I've removed the `exec(compile(...))` magic entirely.

The fixture now does a plain `importlib` load of `cli.py`, and only registers a tiny `langchain_core.tools` stub in `sys.modules` when the real package isn't installed (checked via `importlib.util.find_spec`). So the test still runs offline without ROS2 sourced, but there's no dynamic namespace exec anymore.

5 passed locally. Happy to drop the offline stub and just `skipif` the module instead if you'd prefer keeping the test path even simpler — let me know.

maciejmajek:
@Bartok9 thanks. You can safely assume that langchain-core is installed as this is the core rai library

Bartok9:
Done — thanks @maciejmajek. Dropped the `langchain_core` stub entirely; the fixture now just does a plain `importlib` load of `cli.py` (still bypassing the package `__init__` so ROS2 doesn't need to be sourced). No more dark magic. 5 passed locally.

maciejmajek (approved):

PR #822, opened 2026-07-22
fix(ros2): reject non-positive action server result_timeout

Bartok9:
## Summary

- Reject non-positive / non-numeric `result_timeout` at `ROS2ActionAPI.create_action_server` before constructing the rclpy `ActionServer`.
- `bool` is rejected explicitly (`False` must not pass as `0`).

## Motivation

`result_timeout` controls how long action results are retained. Invalid values (≤0 or non-numeric) should fail at the RAI boundary rather than being forwarded into undefined rclpy behavior.

## Verification

```text
PYTHONPATH=src/rai_core python -m pytest tests/communication_offline/test_action_server_result_timeout.py -q
# 5 passed
```

Offline tests extract and execute the guard from source (no `rclpy` required).

- Did NOT change: default `result_timeout=900`, QoS profiles, send_goal path.

## Notes

Paths are orthogonal to open Bartok9 TF/service work on `connectors/base.py` / `service.py` / `topic.py`.

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h


maciejmajek (changes requested):
Thanks for the contrib! Similarly to the previous PR, the added test will create a burden going  forward. Tests should be easy to understand & maintain. Please consider keeping the path untested or rewrite to minimize complexity.

Bartok9:
Thanks @maciejmajek — understood, and agreed. I've removed the offline test harness in 531b51a; the guard now ships untested to keep the test surface minimal and maintainable. The validation itself (TypeError on non-numeric/bool, ValueError on <=0) is unchanged. Appreciate the review.

maciejmajek (approved):

PR #827, opened 2026-07-23
fix(tools): reject non-positive topic receive/image/TF timeouts

Bartok9:
## Summary
Agent-supplied `timeout_sec` on receive/image/TF tools can be 0/negative/bool/NaN.
Adds `require_positive_timeout` and wires it into ReceiveROS2MessageTool, GetROS2ImageTool, GetROS2TransformTool.

## Testing
- AI-assisted; human-reviewed

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Took this in #872, though as gt=0 on the timeout_sec fields in the args_schema rather than a helper in _run. Pydantic rejects the call before it reaches us and the constraint lands in the tool schema the model sees.

PR #830, opened 2026-07-23
fix(ros2): reject non-positive get_future_result timeout_sec

Bartok9:
## Summary
Fail closed on invalid get_future_result timeouts so waits are never zero/negative/bool/NaN.

## Testing
- AI-assisted; human-reviewed

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
This one already landed with #824, ros_async.py rejects a non-positive timeout_sec on main today.

PR #831, opened 2026-07-23
fix(tools): reject empty ROS topic names on publish/receive/image

Bartok9:
## Summary
Blank topic strings from agents are rejected via require_non_empty_name on publish/receive/image tools.

## Testing
- AI-assisted; human-reviewed

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
In #872 as min_length=1 on the topic fields.

PR #832, opened 2026-07-23
fix(tools): reject empty ROS service name/type on CallROS2ServiceTool

Bartok9:
## Summary
Validate service_name and service_type before writable checks and connector call.

## Testing
- AI-assisted; human-reviewed

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Covered in #872, min_length=1 on service_name and service_type.

PR #836, opened 2026-07-23
fix(nav2): reject non-finite NavigateToPose coordinates

Bartok9:
## Summary
Reject NaN/Inf/bool x/y/z/yaw on NavigateToPoseTool via require_finite_float.

## Testing
- AI-assisted; human-reviewed

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
In #872 as allow_inf_nan=False on the four NavigateToPoseToolInput fields, which is the pydantic equivalent of the finite check.

PR #848, opened 2026-07-25
fix(messages): honor db_path in store_artifacts

Bartok9:
`store_artifacts` built `Path(db_path)` then still opened hardcoded `artifact_database.pkl` in cwd, so custom DB locations never worked.

Use the provided path end-to-end and mkdir parents when creating. Tests cover custom path extend + nested dirs.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Taken as-is in #872.

PR #849, opened 2026-07-25
fix(communication): require positive callback_max_workers

Bartok9:
Validate `BaseConnector(callback_max_workers=...)` is a positive int before constructing ThreadPoolExecutor. Unit tests for reject/accept paths.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Already on main via #824, see base_connector.py:75.

PR #850, opened 2026-07-25
fix(agents): reject empty agents list in runner helpers

Bartok9:
`AgentRunner([])` / `run_agents([])` previously no-op'd or could hang on shutdown wait with nothing to stop. Require a non-empty agents list with a clear ValueError.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
The guard is already there from #824, runner.py raises on an empty list in all three helpers.

PR #851, opened 2026-07-26
fix(ros2): reject non-positive NodeDiscovery period_sec

Bartok9:
NodeDiscovery passed period_sec straight into create_timer. Zero/negative/NaN values fail later or thrash discovery; reject them at init.

Test uses a MagicMock node so ROS does not need to be live.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Closing this one. NodeDiscovery is only constructed internally and nothing in the tree passes period_sec at all, so the default is the only value it ever sees. We are keeping validation at the points where a model or config.toml supplies the value, which is what #872 does.

PR #852, opened 2026-07-26
fix(agents): require positive HRICallbackHandler max_buffer_size

Bartok9:
Chunk aggregation buffer length must be a positive int so streaming never uses a dead zero/negative cap.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Also in from #824, callback.py validates max_buffer_size.

PR #853, opened 2026-07-26
fix(tool_runner): return error ToolMessage for unknown tool names

Bartok9:
Unknown tool_calls raised KeyError and broke the agent loop. Map missing names to a status=error ToolMessage like other tool failures.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
In #872. Worth saying it wasn't actually crashing, the blanket except caught the KeyError, but the message is a lot clearer now.

PR #854, opened 2026-07-26
fix(s2s): validate OpenWakeWord detection threshold

Bartok9:
Confidence threshold must be in (0, 1]. Reject invalid values before downloading/loading the ONNX model.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Rolled into #872, but at the config level. VADConfig and WWConfig validate the threshold in __post_init__ now, since that is where the value comes in from config.toml.

PR #855, opened 2026-07-27
fix(ros2): reject non-positive waiters time_interval

Bartok9:
Poll interval must be > 0. Timeout may still be 0 for wait-forever.

`pytest tests/communication/ros2/test_waiters.py`

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Going to pass on this one. Our own waiter tests call wait_for_ros2_topics with time_interval=0 and time.sleep patched out, which is a reasonable way to run them, so the guard would break them.

PR #856, opened 2026-07-27
fix(s2s): validate SileroVAD detection threshold

Bartok9:
Require Silero VAD `threshold` in (0, 1] and finite before `torch.hub.load`.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Same as #854, handled in #872 on VADConfig rather than inside the model.

PR #857, opened 2026-07-27
fix(s2s): validate SoundDeviceConfig block_size and channels

Bartok9:
Positive ints for block_size; channels and consumer_sampling_rate when set.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
block_size and channels are hardcoded literals in ASRAgent.from_config and the TTS agent, so they never take a value from outside. Closing, the instinct is right but the check needs to sit where a value actually comes in.

PR #858, opened 2026-07-27
fix(s2s): validate BaseTranscriptionModel sample_rate and names

Bartok9:
Guard Whisper/ASR base ctor against non-positive sample rates and blank names.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Same reasoning as #857. sample_rate is passed as 16000 from from_config in every branch, so there is no path for a bad value to arrive. Closing.

PR #859, opened 2026-07-28
fix(frontend): read transcription_model key consistently in ASR config UI

Bartok9:
The ASR vendor selectbox read typo key `transciption_model` while `on_asr_vendor_change` writes `transcription_model`, so a saved vendor could be ignored on reload.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Took the typo fix in #872. Left the test out though, it greps the source file as text and won't survive the next rename.

PR #860, opened 2026-07-28
fix(bench): reject non-positive PlaceCubesTask threshold_distance

Bartok9:
Require finite `threshold_distance` > 0 in PlaceCubesTask so adjacency scoring cannot run with zero/NaN distance.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
The benchmark task kwargs are all set by us when writing a benchmark, so a bad threshold_distance would be a typo caught on the first run. Closing this and the sibling task PRs, not because the checks are wrong but because there is no path for bad input to reach them.

PR #861, opened 2026-07-28
fix(bench): validate GroupObjectsTask threshold and obj_types

Bartok9:
Reject empty `obj_types` and non-positive/non-finite `threshold_distance` in GroupObjectsTask before clustering.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Closing with #860, same reasoning.

PR #862, opened 2026-07-28
fix(s2s): reject non-positive TTSModel sample_rate and channels

Bartok9:
Validate `TTSModel.set_tts_params` so sample_rate and channels must be positive ints before resampling.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Closing alongside #857 and #858. set_params is called from our own connector code with values we control.

PR #864, opened 2026-07-29
fix(bench): validate RotateObjectTask obj_types and quaternion

Bartok9:
Validate obj_types and target quaternion finiteness/non-zero norm on construct.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Closing with #860, same reasoning.

PR #866, opened 2026-07-29
fix(s2s): validate VADConfig and WWConfig thresholds and grace

Bartok9:
VADConfig/WWConfig __post_init__ guards: threshold in (0,1], positive finite silence_grace_period.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
This is the one I kept from the threshold set, in #872. Both VADConfig and WWConfig validate now, matching what TranscribeConfig already did.

PR #868, opened 2026-07-31
fix(bench): reject empty Scenario scene_config_path

Bartok9:
Reject empty/whitespace Scenario.scene_config_path so experiment logging cannot store blank config paths.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Closing with the other bench PRs. 116 lines to check that a path string is not empty is more than we want to carry for something that fails immediately and visibly when it does happen.

PR #869, opened 2026-08-01
fix(bench): reject negative TaskArgs.extra_tool_calls

Bartok9:
Negative `extra_tool_calls` could inflate tool-calling pass budgets. Validate `>= 0` and reject bool before coercion.

AI-assisted; human-reviewed.

Claim: bartok
Operator: bartok
Campaign: aerial-drone


maciejmajek:
Closing, though there was more behind this one than the rest. A negative extra_tool_calls really does drop max_tool_calls_number below required_calls and skew results instead of erroring. If we take it, it will be ge=0 on the pydantic field rather than a validator plus a test module, so I would rather fold it in next time that file is open.

